PR #24, opened 2016-11-02
Fixed npm command in windows

### LINKED ISSUE #15: Add Windows support
Currently, the `npm` command is not found on windows.

DylanArnold:
Fixes #15

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Commands.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Commands.scala
@@ -3,6 +3,10 @@ package scalajsbundler
 import sbt._
 
 object Commands {
+  private val npm = sys.props("os.name").toLowerCase match {
+    case os if os.contains("win") ⇒ "cmd /c npm"
+    case _ ⇒ "npm"
+  }
 
   /**
     * Runs the `npm update` command
@@ -11,7 +15,7 @@ object Commands {
     */
   def npmUpdate(cwd: File, log: Logger): Unit = {
     log.info("Updating NPM dependencies")
-    run("npm update", cwd, log)
+    run(s"$npm update", cwd, log)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/Commands.scala
```diff
@@ -3,6 +3,10 @@ package scalajsbundler
 import sbt._
 
 object Commands {
+  private val npm = sys.props("os.name").toLowerCase match {
+    case os if os.contains("win") ⇒ "cmd /c npm"
```
julienrf: Can you explain what does that mean?

DylanArnold: It's testing what the current operating system is. Then it uses the built in `cmd` executable in windows to execute the `npm` batch file.

I don't really know why but it seems as though executing a shell command doesn't pick up windows batch files, only executables.

So if it's windows it executes `cmd /c npm update` etc, if it's some other os, just `npm update`. (See the usage of the npm val in string interpolation below).

I got the sys.props trick by looking at the sbt-native-packager source code.

Not really certain if there is better approach but this seems to work.

DylanArnold:
I don't think you need to update the usage of the node command because it is a windows executable rather than a batch file.

So far this seems to be working. This is my first time using scalajs-bundler so I haven't completely got a working project yet, however I noticed that the bundler js files are being created, so everything appears fine now.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Commands.scala
DylanArnold: Oops I just noticed my unicode arrows ⇒. I can update those to => ?

DylanArnold:
What's needed to test the node command?

julienrf:
Use the last version and run the `fastOptJS::webpack` sbt command.

julienrf:
OK, thanks again @DylanArnold for your work on this, can you confirm me that you can run the `fastOptJS::webpack` command with the last version? Then I think this is good to merge!

DylanArnold:
@julienrf No problem. Thanks for all the hard work you put into this project :) 

I just tested `fastOptJS::webpack` and it executes with success.

julienrf:
Great!

PR #28, opened 2016-11-03
Ignore IntelliJ .idea directory

DylanArnold:


DylanArnold:
Hmm I just noticed the ugly Merge remote-tracking branch commit that Github desktop added. Sorry, I wonder if I can abort / remove that.

julienrf:
Hi Dylan,

Thanks for the effort but I think I prefer users to use a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for this kind of user sepcific settings. Does that make sense to you too?

Also, for your next PRs, It would be better to create a distinct branch on your fork so that I don’t have to merge your merge commits. (Something like what is described [here](https://github.com/scala-js/scala-js/blob/master/CONTRIBUTING.md#general-workflow) ; at some point I should add such kind of description to my contributing.md file)

DylanArnold:
Okay fair enough, that makes some sense. I'll go with that :) Thanks for the link!

PR #64, opened 2016-12-13
Webpack2 support

julienrf:


julienrf:
Switching to yarn instead of npm to fetch the dependencies fixed the CI issue… It must be related to some locked file in the Windows file system.

PR #77, opened 2016-12-23
Extending the cookbook by a small note

Sebruck:
This gave me some headache and I had to look into the code to understand what's happening.

The **goal** of the PR is that future reader will not fall into the same hole ;-)

PR #78, opened 2016-12-23
Extending the cookbook by a small note

Sebruck:
This is a followup of https://github.com/scalacenter/scalajs-bundler/pull/77 

I created a new branch so that i get a clean history because i had some problems while rebasing.


julienrf:
Thanks! The doc is updated.

PR #81, opened 2016-12-27
Disabled reload workflow by default because it causes confusion.

### LINKED ISSUE #60: Disable the reload workflow by default
(Also see #58 for some more context)

The “reload workflow” is a special mode that replaces the `webpack` task implementation with a different one, that does not use the webpack program to process the output of the Scala.js compilation. It has been implemented to get a faster “change source and reload application” workflow but a drawback is that it ignores any user supplied webpack configuration file. This behavior might be surprising so it is probably better to not enable it by default.

Sebruck:
This PR fixes #60

### DIFF
--- a/manual/src/ornate/reference.md
+++ b/manual/src/ornate/reference.md
@@ -68,6 +68,25 @@ useYarn := true
 
 The `yarn` command must be available in the host platform.
 
+### Reload Workflow {#reload-workflow}
+
+Each time you change something in your application source code and compile the project, Scala.js emits a new .js file 
+that can weigh several MBs (in fastOptJS mode) if your application is large. Turning this huge CommonJS module into 
+code executable by web browsers takes time. This can be a problem if you rely on a “live-reloading” workflow 
+(like the one of Play framework, for instance), because the reloading time can go up to 30 seconds.
+
+You can get a faster “change source and reload application” workflow by setting the `enableReloadWorkflow` 
+key to `true`.
+
+The reload workflow replaces the `fastOptJS::webpack` task implementation with a different one, that does not use 
+webpack to process the output of the Scala.js compilation.  Instead, it pre-bundles the modules imported by your 
+application and exposes them to the global namespace. Since these dependencies can then be resolved from the global 
+namespace, the output of Scala.js is just concatenated after the contents of the pre-bundling process.
+
+> {.note}
+> As soon as `enableReloadWorkflow` is true `fastOptJS::webpack` does **not** use webpack and therefore 
+> the custom webpack configuration file is ignored.
+
 ### Tasks and Settings {#tasks-and-settings}
 
 The tasks and settings that control the plugin are documented in the API documentation
--- a/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace/build.sbt
+++ b/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace/build.sbt
@@ -22,8 +22,6 @@ npmDevDependencies in Compile += "expose-loader" -> "0.7.1"
 // as expected by the scalajs-react facade
 webpackConfigFile := Some(baseDirectory.value / "webpack.config.js")
 
-// Needed to make scalajs-bundler actually use the custom config file in fastOptJS
-enableReloadWorkflow := false
 //#relevant-settings
 
 // (Used by tests only) checks that a HTML can be loaded (and that its JavaScript can be executed) without errors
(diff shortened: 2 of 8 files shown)


manual/src/ornate/reference.md
```diff
@@ -68,6 +68,17 @@ useYarn := true
...
+
+To get a faster "change source and reload application" workflow enable the setting `enableReloadWorkflow`.
+
+The reload workflow is a mode that replaces the `webpack` task implementation with a different one,
+that does not use webpack to process the output of the Scala.js compilation. 
+
+> {.note}
+> As soon as `enableReloadWorkflow` is true `fastOptJS::webpack` does **not** use webpack and therefore 
+> the custom webpack configuration file is ignored.
+
```
julienrf: Could you also add an entry in the cookbook? Something like the following:

> # How to improve the performance of the bundling process? {#performance}
> 
> You can enable the [reload workflow](reference#reload-workflow) and disable source maps:
> 
> ~~~ scala
> enableReloadWorkflow := true
> emitSourceMaps := false
> ~~~

manual/src/ornate/reference.md
```diff
@@ -68,6 +68,17 @@ useYarn := true
 
 The `yarn` command must be available in the host platform.
 
+### Reload Workflow {#reload-workflow}
+
+To get a faster "change source and reload application" workflow enable the setting `enableReloadWorkflow`.
```
julienrf: And I would also start with the following paragraph:

> Each time you change something in your application source code and compile the project, Scala.js emits a new .js file that can weigh several MBs (in `fastOptJS` mode) if your application is large. Turning this huge CommonJS module into code executable by web browsers takes time. This can be a problem if you rely on a “live-reloading” workflow (like the one Play framework, for instance), because the reloading time can go up to 30 seconds.

manual/src/ornate/reference.md
```diff
@@ -68,6 +68,17 @@ useYarn := true
...
+
+The reload workflow is a mode that replaces the `webpack` task implementation with a different one,
+that does not use webpack to process the output of the Scala.js compilation.
```
julienrf: I would also add:

> Instead, it pre-bundles the modules imported by your application and exposes them to the global namespace. Since these dependencies can then be resolved from the global namespace, the output of Scala.js is just concatenated after the contents of the pre-bundling process.

julienrf:
test this please

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace/build.sbt
```diff
@@ -22,8 +22,6 @@ npmDevDependencies in Compile += "expose-loader" -> "0.7.1"
 // as expected by the scalajs-react facade
 webpackConfigFile := Some(baseDirectory.value / "webpack.config.js")
 
-// Needed to make scalajs-bundler actually use the custom config file in fastOptJS
-enableReloadWorkflow := false
```
Sebruck: Why didn't the `static` test fail after my changes? Shouldn't it fail because it now uses https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/static/prod.webpack.config.js the UglifyJsPlugin  hopefully resulting in a different file size?

julienrf: Because the reload workflow is always disabled for `fullOptJS`. We should add a note about that.

manual/src/ornate/reference.md
```diff
@@ -68,6 +68,17 @@ useYarn := true
 
 The `yarn` command must be available in the host platform.
 
+### Reload Workflow {#reload-workflow}
+
+To get a faster "change source and reload application" workflow enable the setting `enableReloadWorkflow`.
+
+The reload workflow is a mode that replaces the `webpack` task implementation with a different one,
```
julienrf: I would say “replaces the `fastOptJS::webpak` task implementation”

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace/build.sbt
Sebruck: ahh ok, i oversaw that it's doing full instead of fast

Sebruck:
@julienrf  Thx for the comments, i completed all of them.

julienrf:
Great :+1:

PR #82, opened 2017-01-04
Make scalaJSBundlerPackageJson accessible

muuki88:
The purpose of this PR is to make the `scalaJSBundlerPackageJson` available so anyone can
override to create a custom `package.json`.

## Use Case

I'm trying to build a _sbt-native-packager_ archetype that bundles a complete node application. It includes the complete `node_modules` folder to run the application. In the _production_ build I don't want to include all the dev dependencies, so I want to remove the from the `package.json`. For this reason I want to create a custom `package.json`.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -230,10 +230,9 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
     val useYarn: SettingKey[Boolean] =
       settingKey[Boolean]("Whether to use yarn for updates")
 
+    val scalaJSBundlerPackageJson =
+      TaskKey[File]("scalaJSBundlerPackageJson", "Write a package.json file defining the NPM dependencies of project")
   }
-
-  private val scalaJSBundlerPackageJson =
-    TaskKey[File]("scalaJSBundlerPackageJson", "Write a package.json file defining the NPM dependencies of project", KeyRanks.Invisible)
 
   private val scalaJSBundlerWebpackConfig =
     TaskKey[File]("scalaJSBundlerWebpackConfig", "Write the webpack configuration file", KeyRanks.Invisible)


julienrf:
Hi, I think the common way to publish Node.js app is to publish the `package.json` file rather than the `node_modules` directory.

muuki88:
Publishing the `package.json` means that you have to run `npm install` or `yarn` before running your application, right? I have no real experience in packaging node applications, but wanted to create a similar experience as with the `JavaAppPackaging` archetype in _sbt-native-packager_.

`JavaAppPackaging` creates the following output

```
bin/
    start
    start.bat
lib/
   .. <library dependencies> 
conf/
   .. <optional config files>
```

I want to replicate this output so that `lib -> node_modules`.

julienrf:
OK, I understand.

Did you try to publish a bundle instead of a .js file + the `node_modules` directory?

muuki88:
> Did you try to publish a bundle instead of a .js file + the node_modules directory?

Yes, but it didn't worked. I guess webpack doesn't export the modules like node requires. The example project is here: https://github.com/muuki88/node-scalajs

julienrf:
Did you use the `target` option? (see here: http://webpack.github.io/docs/configuration.html#target)

muuki88:
Yeah, I had this in my [first working version](https://github.com/muuki88/node-scalajs/tree/8b1e1b63c03f139740256cfca747b8b7a4cb8b46), but without any effect.

julienrf:
Oh, indeed it has only an effect on dynamically loaded chunks (but code splitting is not supported yet by scalajs-bundler, so the `target: 'node'` option has no effect).

However, the process you describe [here](https://github.com/muuki88/node-scalajs/tree/8b1e1b63c03f139740256cfca747b8b7a4cb8b46#run) works for me:

~~~ sh
$ node target/universal/stage/node-js-app.js    
hello world v4 14dcf6a7-1049-4569-9d9f-cd59c886d381
hello world v1 295d3740-d26a-11e6-8b80-dd1cada9bd4a
~~~

muuki88:
Yes. It does work, but the `node_modules` folder contains all the `dev_dependencies`. I would like to remove these ;)

julienrf:
Actually, the `target: 'node'` option *has* an effect, but for whatever reason your custom webpack configuration file is ignored that’s why it has no effect in your case.

Once you’ll solve this problem then your bundle will be compatible with Node.js and you will not need the `node_modules` folder at all (since all the used dependencies will be included in the bundle).

julienrf:
Basically, I think your plugin requires a lot *less* machinery: just include the output of the `webpack` task in the `mappings`.

muuki88:
Thanks for taking a close look :) 
I will try again and see if I can make it work with webpack.

Still, WDYT of making this task accessible?

julienrf:
I prefer keeping most of the things private, to avoid having to deal with backward compatibility in the future.

PR #90, opened 2017-01-11
Add support for testHtml{Full,Fast}Opt

dispalt:


julienrf:
After thinking a little bit more I think the solution I suggested was wrong. I’ll have a deeper look tomorrow.

dispalt:
I think I fixed it, take a look, you can explicitly override the scope.

julienrf:
Unfortunately, I think the solution **I** gave you at the beginning was wrong because the HTML test runner does not want the *application* bundle but the *test* bundle, so we should not use the `webpack` command.

We generate the test bundle at this place: https://github.com/scalacenter/scalajs-bundler/blob/efb3ee90cfc9047cd17d960db16afbedc5e081b7/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala#L368

As you can see we just append `-bundle` to the test file name. So, your initial solution was actually the correct one!

You can check if the generated HTML runner does work by adding the following command in the `test` file:

~~~
> html target/scala-2.12/static-fastopt-test.html
~~~

dispalt:
Rebased and addressed the comments.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/static/test
```diff
@@ -32,4 +32,11 @@ $ absent target/scala-2.11/scalajs-bundler/main/static-fastopt-bundle.js.map
...
-> test
\ No newline at end of file
+> test
+
+# Generate the html test report
+
+$ absent target/scala-2.11/static-fastopt-test.html
+> testHtmlFastOpt
+$ exists target/scala-2.11/static-fastopt-test.html
+> html target/scala-2.11/static-fastopt-test.html
```
julienrf: Could you please also add the same for `testHtmlFullOpt`?

julienrf:
test this please

dispalt:
testHtmlFullOpt is not working, investigating

julienrf:
Hey @dispalt I pushed a commit that fixes the test. However, the test do not pass, probably because I’m forcing the use of `fastOptJS` somewhere in the ScalaJSBundlerPlugin whereas it should be `fullOptJS`.

dispalt:
It looks like its failing on expected size

julienrf:
Yes, there is also this error but even though we fix it the test for `testHtmlFullOpt` fails. I will fix the size error in `master`.

dispalt:
Yeah I see it now, Ill take a look

dispalt:
I think this is changing per 1.0 comments, closing.

sjrd:
Indeed. The changes in 1.0 were even partly motivated by the issues encountered in this PR ;)

PR #94, opened 2017-01-17
Fix #93

### LINKED ISSUE #93: Improve caching behavior for tasks
webpack task
--------------------
Current implementation of the `webpack` task only actually launches webpack if one the following changes:
1. Generated configuration files (`package.json`, `scalajs.webpack.config.js`).
2. Custom webpack configuration file (if any).
3. Entries generated by `scalajs`.

This is a major inconvenience in the following use cases:
1. If the custom config imports another file (for example if few configs reuse some base settings), webpack won't be launched if only the imported file has changed.
2. If the code imports some unmanaged resource (script or something else), webpack won't be launched.

There is currently no satisfactory way to circumvent this behavior. Here are a few ideas:
1. Expose a setting to allow appending to the list of files being checked. 
2. Expose the same functionality without caching as a separate task.
3. Just disable caching for this task.

package.json generation
--------------------------------
Current implementation only regenerates `package.json` if running with changed sbt `Configuration` (`Compile`, `Test`, etc.). This means that changing `npmDependencies`/`npmDevDependencies` won't trigger `package.json` update.

Possible solutions:
1. Append these settings to the hash in `PackageJsonTasks.scala`.
2. Hash the generated data instead of just using configuration name.

nigredo-tori:
Implements changes proposed in #93:
- Fix `scalaJSBundlerPackageJson` caching behavior.
- Expose `webpack` task caching settings.

I had to use a slightly more involved API for monitored files (`webpackMonitoredDirectories` and `webpackMonitoredIncludeFilter` settings, `webpackMonitoredFiles` task). A single file list setting is not flexible enough, and a file list task that is supposed to be overridden is overkill in most cases.

Fixes #93

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -208,30 +208,16 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
       * Additional directories, monitored for webpack launch.
       *
       * Changes to files in these directories that match
-      * `webpackMonitoredIncludeFilter` enable webpack launch in
-      * `webpack` task.
+      * `includeFilter` scoped to `webpackMonitoredFiles` enable
+      *  webpack launch in `webpack` task.
       *
       * Defaults to an empty `Seq`.
       * 
       * @group settings
-      * @see [[webpackMonitoredIncludeFilter]]
+      * @see [[webpackMonitoredFiles]]
       */
     val webpackMonitoredDirectories: SettingKey[Seq[File]] =
       settingKey[Seq[File]]("Directories, monitored for webpack launch")
-
-    /**
-      * Filter for files, monitored for webpack launch.
-      * 
-      * Changes to files matching this filter in `webpackMonitoredDirectories`
-      * enable webpack launch in `webpack` task.
-      * 
-      * Defaults to `AllPassFilter`
-      * 
-      * @group settings
-      * @see [[webpackMonitoredDirectories]]
-      */
-    val webpackMonitoredIncludeFilter: SettingKey[FileFilter] =
-      settingKey[FileFilter]("Filter for files, monitored for webpack launch")
 
     /**
       * List of files, monitored for webpack launch.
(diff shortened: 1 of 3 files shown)


julienrf:
test this please

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -201,6 +205,54 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
+      * Changes to files matching this filter in `webpackMonitoredDirectories`
+      * enable webpack launch in `webpack` task.
+      * 
+      * Defaults to `AllPassFilter`
+      * 
+      * @group settings
+      * @see [[webpackMonitoredDirectories]]
+      */
+    val webpackMonitoredIncludeFilter: SettingKey[FileFilter] =
+      settingKey[FileFilter]("Filter for files, monitored for webpack launch")
```
julienrf: Why not reuse the existing `includeFilter` key, scoped to the `webpackMonitoredFiles` task?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -450,8 +507,27 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
+
+        val additionalFiles = dirs.flatMap(
+          dir => (dir ** filter).get
+        )
+
+        packageJsonFile +:
+          generatedWebpackConfigFile +:
+          customWebpackConfigFile ++:
+          entries.map(_._2) ++:
+          additionalFiles
```
julienrf: I think you should also add `stageTask.value.data`, which contains the .js code produced by Scala.js (the entries contain only the launcher files, which depend on them). I don’t know how we could automate the resolution of the dependencies among .js files…

nigredo-tori: Will do. I seem to have lost this one in the transition.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
nigredo-tori: You're right, of course. Scoped `includeFilter` is more idiomatic. Will change this shortly.

julienrf:
Awesome, thanks!

PR #96, opened 2017-01-18
Add webpack-dev-server wrapper

nigredo-tori:
A simple wrapper that runs `webpack-dev-server` in a background thread.
Possible problems:
1. ~~Current implementation exposes tasks/settings on project level, but uses some settings from `(Compile, fastOptJS)`. I can't see a use case for multiple dev servers in one project, though. Is there a nicer way to scope this?~~ Moved the tasks to stage level for consistency/flexibility. Settings are still at the project level, which is more convenient and doesn't cost us flexibility.
2. Currently `startWebpackDevServer` task partially duplicates file layout logic of `Webpack`. This should probably be refactored. Ideally, I'd like to see paths to generated files that are calculated directly from settings (copied/generated webpack configs, for example) to be settings, too, rather than results of tasks.
3. ~~The provided test uses some platform-specific tools (which won't work out of the box on Windows), and assumes there are no other `node` processes running. Is there any way around this?~~ Replaced `pgrep` checks with http queries.


julienrf:
Hi, thanks for working on this. However I must confess that I’m not sure which problem this PR solves. Can you elaborate on that? What would be the difference with, for instance running the `~fastOptJS::webpack` task?

julienrf:
test this please

nigredo-tori:
I've fixed the test, and opted to move the tasks to stage level. As for the paths refactoring - I don't think it should be a part of this PR, but rather a separate issue.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-dev-server/build.sbt
```diff
@@ -0,0 +1,29 @@
...
+scalaVersion := "2.11.8"
+
+// Use a custom config file
+webpackConfigFile := Some(baseDirectory.value / "webpack.config.js")
+
+(npmDevDependencies in Compile) += ("html-webpack-plugin" -> "2.24.1")
+
+webpackDevServerPort := 7357
+
+//#relevant-settings
```
julienrf: This line is irrelevant

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-dev-server/test
```diff
@@ -0,0 +1,17 @@
...
+
+# the server should properly stop
+> fastOptJS::stopWebpackDevServer
+-> html index.html
+
+# the server should properly terminate on exit
+> fastOptJS::startWebpackDevServer
+$ sleep 3000
+> html index.html
+-> exit
```
julienrf: Why is it prefixed by `-`?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
```diff
@@ -42,7 +42,8 @@ object PackageJson {
       ) ++ Seq(
         "webpack" -> webpackVersion,
         "concat-with-sourcemaps" -> "1.0.4", // Used by the reload workflow
-        "source-map-loader" -> "0.1.5" // Used by webpack when emitSourceMaps is enabled
+        "source-map-loader" -> "0.1.5", // Used by webpack when emitSourceMaps is enabled
+        "webpack-dev-server" -> "1.16.2"
```
julienrf: I would rather install it independently, like we do for jsdom, instead of adding it to the project’s dependencies.

manual/src/ornate/cookbook.md
```diff
@@ -190,3 +190,33 @@ You can enable the [reload workflow](reference.md#reload-workflow) and disable s
...
+    ~~~
+    > fastOptJS::stopWebpackDevServer
+    ~~~
+
+Additional arguments can be passed to webpack-dev-server via `webpackDevServerExtraArgs`
+setting. For example, you can add the following to your `build.sbt` to make your page
+reload on every change:
+~~~
+webpackDevServerExtraArgs := Seq("--inline")
+~~~
```
julienrf: Did you check that this markdown renders corretly? (Github displays it incorrectly)

nigredo-tori: It renders correctly with `manual/previewSite`. And Github doesn't like the markdown in master branch, either. I didn't spend much time trying to reconcile the two, though.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-dev-server/test
nigredo-tori: Because `scripted` thinks that this is an error (probably because SBT doesn't return back to prompt after this one). I found no other way to test the exit hook, though (`reload``isn't the same, since it runs unload hooks, and `exit` doesn't).

julienrf:
test this please

julienrf:
Thanks!

PR #97, opened 2017-01-21
If useYarn is true, then use it to install the webpack dev server

dispalt:


julienrf:
test this please

julienrf:
Good catch!

We should do the same with jsdom installation, actually.

julienrf:
Thanks :)

Issue #112, opened 2017-03-19
webpack task caching ignores changes to custom webpack config files


japgolly:
Say you have

```scala
webpackConfigFile in fastOptJS := Some(baseDirectory.value / "webpack-dev.js"),
webpackConfigFile in fullOptJS := Some(baseDirectory.value / "webpack-prod.js"),
```

Then:
1. `fastOptJS::webpack`
2. Modify `webpack-dev.js`.
3. `fastOptJS::webpack` ← should re-run webpack but doesn't.

danxmoran:
@japgolly / @julienrf is this still a problem in 0.7? I've been trying to reproduce in a minimal scripted test without any luck.

julienrf:
I think this has been fixed by 32d852d533f8832f78d5152425ad97453fb40e83

Issue #83, opened 2017-01-05
The test framework ignores any custom webpack configuration file


julienrf:

julienrf:
The problem is that the custom configuration file might define several bundles (with arbitrary names), but the test framework expects to get exactly one bundle (whose code exposes some specific properties to the global namespace).

PR #119, opened 2017-04-05
Support for custom webpack config in reload workflow

jeroentervoorde:
I implemented this when i ran into an issue using leaflet assets which contain images and require loader configuration (according to my very minimal webpack knowledge..). This failed when using reload workflow. The bundled js is now created using the optional webpack configuration set by "webpackConfigFile in webpackReload"

To be able to scope the setting i added the fastOptJS::webpackReload task which does the same thing as fastOptJS::webpack with reloadWorkflow enabled.

The reloadWorkflow setting still works and also supports the webpackConfigFile setting.

If the approach is ok i'll try to add a test.

### DIFF
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -72,9 +72,9 @@ In addition to the configured webpack config file, all .js files in the project 
 from the various configuration files.
 
 Here are the steps to share the loader configuration among your prod and dev config files. This
-uses webpack-merge for convenience the same result could be accomplished using plain js only.
+uses webpack-merge for convenience. The same result could be accomplished using plain js only.
 
-1. Put configuration it in a common.webpack.config.js file:
+1. Put configuration in a common.webpack.config.js file:
 
 ~~~ javascript
 module.exports = {
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -53,6 +53,63 @@ More fine-grained control over the list of monitored files is possible by overri
 You can find a working example of custom configuration file
 [here](https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/static/prod.webpack.config.js).
 
+It is also possible to configure a webpack config file to be used in reload workflow and when running the tests.
+This configuration may not contain `entry` and `output` configuration but can be used to configure loaders etc.
+
+These configuration files are configured using `webpackConfigFile in reloadTask` or `webpackConfigFile in Test`.
+For example:
+
+~~~ scala
+webpackConfigFile in webpackReload := Some(baseDirectory.value / "common.webpack.config.js")
+
+webpackConfigFile in Test := Some(baseDirectory.value / "common.webpack.config.js")
+~~~
+
+## Sharing webpack configuration among configuration files {#shared-config}
+
+In addition to the configured webpack config file, all .js files in the project base directory
+(as configured using the `webpackResources` setting) are copied to the target directory so they can be imported
+from the various configuration files.
+
+Here are the steps to share the loader configuration among your prod and dev config files. This
+uses webpack-merge for convenience the same result could be accomplished using plain js only.
+
+1. Put configuration it in a common.webpack.config.js file:
+
+~~~ javascript
+module.exports = {
+  module: {
+    loaders: [
+        ...
+    ],
+    rules: [
+        ...
+    ]
+  }
+}
+~~~
+
+2. Add webpack-merge to your npmDevDependencies:
+
+~~~
+npmDevDependencies in Compile += "webpack-merge" -> "4.1.0"
+~~~
+
+3. Merge in the common configuration in your dev.webpack.js file:
+
+~~~ javascript
+var merge = require("webpack-merge")
+var commonConfig = require("./common.webpack.config.js")
+
+module.exports = merge(commonConfig, {
+    ...
+})
+~~~
+
+You can find a working example of a project using a shared configuration file
+[here](https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig).
+
+
 ## How to use npm modules from Scala code? {#facade}
 
 Once you have [added npm dependencies](getting-started.md) to the packages you are interested
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
@@ -127,6 +127,7 @@ object ReloadWorkflow {
     workingDir: File,
     entryPoint: File,
     bundleFile: File,
+    customWebpackConfigFile: Option[File],
     logger: Logger
   ): Unit = {
     logger.info("Pre-bundling dependencies")
@@ -139,7 +140,13 @@ object ReloadWorkflow {
       )
     IO.write(entryPoint, depsFileContent.show)
 
-    Webpack.run(entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
+    customWebpackConfigFile match {
+      case Some(configFile) =>
+        Webpack.run("--config", configFile.getAbsolutePath, entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
+      case None =>
+        Webpack.run(entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
+
+    }
 
     ()
   }
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
@@ -128,9 +128,22 @@ object ReloadWorkflow {
     entryPoint: File,
     bundleFile: File,
     customWebpackConfigFile: Option[File],
+    sharedWebpackConfigFiles: Seq[File],
     logger: Logger
   ): Unit = {
     logger.info("Pre-bundling dependencies")
+
+    val customConfigFile =
+      customWebpackConfigFile.map { file =>
+        val configFileCopy = workingDir / file.name
+        IO.copyFile(file, configFileCopy)
+        configFileCopy
+      }
+
+    sharedWebpackConfigFiles.foreach { file =>
+      val sharedConfigFileCopy = workingDir / file.name
+      IO.copyFile(file, sharedConfigFileCopy)
+    }
 
     val depsFileContent =
       JS.block(
@@ -140,12 +153,11 @@ object ReloadWorkflow {
       )
     IO.write(entryPoint, depsFileContent.show)
 
-    customWebpackConfigFile match {
+    customConfigFile match {
       case Some(configFile) =>
         Webpack.run("--config", configFile.getAbsolutePath, entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
       case None =>
         Webpack.run(entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
-
     }
 
     ()
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
@@ -8,6 +8,7 @@ import org.scalajs.sbtplugin.Loggers
 import org.scalajs.sbtplugin.ScalaJSPlugin.AutoImport._
 import sbt._
 
+import scalajsbundler.Webpack.copyToWorkingDir
 import scalajsbundler.util.{Commands, JS}
 
 /**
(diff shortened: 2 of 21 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/ReloadWorkflow.scala
```diff
@@ -139,7 +140,13 @@ object ReloadWorkflow {
       )
     IO.write(entryPoint, depsFileContent.show)
 
-    Webpack.run(entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
+    customWebpackConfigFile match {
+      case Some(configFile) =>
+        Webpack.run("--config", configFile.getAbsolutePath, entryPoint.absolutePath, bundleFile.absolutePath)(workingDir, logger)
```
julienrf: Does that mean that webpack still uses the given `entryPoint` and `bundleFile` instead of using those given in the configuration file?

jeroentervoorde: No, it doesn't. I created a stripped down version of my webpack config containing only loader and rules configuration:

{{{
module.exports = {
  module: {
    loaders: [
     {
        test: /\.(jpe?g|png|gif|svg)$/i,
        loaders: [
          'file?hash=sha512&digest=hex&name=[hash].[ext]',
          'image-webpack?bypassOnDebug&optimizationLevel=7&interlaced=false'
        ]
      }
    ],
    rules: [
      {
        test: /\.css$/,
        use: [ 'style-loader', 'css-loader' ]
      },
      {
        test: /\.(png|jpg|gif|svg|eot|ttf|woff|woff2)$/,
        loader: 'url-loader',
        options: {
          limit: 10000
        }
      }
    ]
  }
}
}}}

julienrf: That’s great to be able to use a custom configuration file as you do, but I think that’s not good if we have to duplicate parts of the config between the “reload workflow” and the usual workflow.

I think that would be easy to share the same config, though. For instance, you could have a `webpack-module.js` file containing the code you pasted. Then we could have a `webpack.config.js` file that would import this one… That’s a bit complex but that would work…

Cool, could you add a test that does that?

jeroentervoorde: That would work i guess. One problem is that the custom webpack file is copied to the target dir which. Therefore the 'module' config needs to be imported using an absolute path.

Maybe a better approach would be to actually add a config setting for the modules only. This can than be used for the reload workflow but also be automatically included in the generated webpack config. A custom config file would still need to manually include the module config but it would be located in the same directory where the custom file is copied.

I imported my custom config using 
var modulesConfig = require('/home/.../project/js/webpack-reload.config.js')

modulesConfig.modules should than be merged with the configuration.

Is this the way to do it?

julienrf: > I could just copy the module config to the target dir as well of course

That’s what we do with the custom webpack config file.

> Maybe a better approach would be to actually add a config setting for the module-config only.

I’m not sure about that. This configuration entry can contain arbitrary JavaScript expressions (for instance a JavaScript `RegExp`). Providing support for that would require some effort (unless we make it possible to write the configuration module in Scala.js). That’s why so far I tried to make as few configuration settings as possible available from scalajs-bundler: for other use cases you should fallback to a plain JavaScript config file.

jeroentervoorde: Actually i meant a setting for the module config file that would replace the "webpackConfigFile in reloadWorkflow" i use now. This file can than be copied to target and used in reloadWorkflow and included in the generated webpack config (using require and maybe some js for merging the settings). Because it is copied it can also be included in a custom config using a relative filename.

jeroentervoorde:
I added a setting webpackSharedConfigFiles which contains a list of files copied to the webpack working dir. This file can than be included in a webpackCustomConfig file.

A test (sharedconfig) was added as a working example. While writing the test i noticed that the test runner also did not support a custom webpack file so i added this ("webpackCustomConfig in test") which can normally point to the same file as "webpackCustomConfig in webpackReload"

In the example the custom config file and the shared config could have been 1 file. In this case it would have been copied twice but i assume this is not a problem.

The pr needs more scaladoc. I'll add that if we agree on this implementation.

julienrf:
Hi, I think this is great! (and that would also probably solve #83)

However, I would propose the following change. Instead of having this `sharedWebpackConfigFiles` setting I would simply copy all the .js files that are in the `baseDirectory`. Actually, this should be configurable, so we would have a `webpackResources` setting of type `PathFinder` that would default to `baseDirectory.value ** "*.js"` (thus, that would include both the config files and the shared config files). Then we would copy all the `webpackResources` into the target directory before invoking webpack. What do you think? (Also, that might help us to fix #112)

If you could update the PR with that that would be awesome! (Then I might add some minor comments, but I would first like to validate the general solution)

jeroentervoorde:
Sure, i'll update it this week. My plan is to append the webpackConfigFile to the files returned by webpackResources as it may not be location in the project root. For #112 we could than hash over all these config files. Another option would be to enforce that webpackConfigFile is part of webpackResources but that's a bit more limiting.

The idea I had for the sharedWebpackConfigFiles was that it could be used as to implement merging in the generated config files as well (in a new PR). This can still be done but would still require a setting besides webpackResources.

jeroentervoorde:
Hi! Updated again. sharedWebpackConfigFiles -> webpackResources (using baseDirectory.value * "*.js" as default so only the files in the project root are copied)

webpackCustomConfig is copied as well if it wasn't by webpackResources.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
```diff
@@ -6,6 +6,14 @@ import scalajsbundler.util.{Commands, JS}
 
 object Webpack {
 
+  def copyToWorkingDir(targetDir: File)(file: File): File = {
+    val copy = targetDir / file.name
+    if (copy.getAbsolutePath != file.getAbsolutePath && !copy.exists()) {
```
julienrf: Why are these tests necessary? Does that happen to have the same source and destination? In such a case maybe we should let the IO.copyFile method throw an error. Also, the second test will make it impossible to update a file that has been copied and then changed afterwards. Unless I’m missing something?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -587,20 +588,23 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
 
                 val customWebpackConfigFile = (webpackConfigFile in test).value
-                val sharedWebpackConfigFiles = webpackSharedConfigFiles.value
+                val sharedWebpackConfigFiles = webpackResources.value.get
+
+                def copyToWorkingDir(file: File): File = {
+                  val copy = targetDir / file.name
+                  IO.copyFile(file, copy)
+                  copy
+                }
```
julienrf: Why can’t you reuse the one defined in the `Webpack` object?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -465,6 +502,10 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
         else webpackTask(fastOptJS)
       }.value,
 
+      webpackReload in fastOptJS := Def.taskDyn {
+        ReloadWorkflowTasks.webpackTask(configuration.value, fastOptJS)
+      }.value,
```
julienrf: Why is this a `taskDyn` rather than a `task`?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -546,6 +587,24 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
                 val sjsOutputName = sjsOutput.name.stripSuffix(".js")
                 val bundle = targetDir / s"$sjsOutputName-bundle.js"
 
+                val customWebpackConfigFile = (webpackConfigFile in test).value
```
julienrf: Why not use the `Test` configuration instead of the `test` task?

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/build.sbt
```diff
@@ -0,0 +1,55 @@
...
+
+// Use a different Webpack configuration file for production
+webpackConfigFile in fullOptJS := Some(baseDirectory.value / "prod.webpack.config.js")
+
+webpackConfigFile in test := Some(baseDirectory.value / "reload.webpack.config.js")
+
+// Use a different Webpack configuration file for reload workflow
+webpackConfigFile in webpackReload := Some(baseDirectory.value / "reload.webpack.config.js")
+
+webpackSharedConfigFiles := Seq(baseDirectory.value / "common.webpack.config.js")
```
julienrf: Why not leave the default setting?

julienrf:
test this please

dispalt:
So just to understand this, does this allow us to have a webpack conf and still use the fast reload workflow?

julienrf:
test this please

julienrf:
@dispalt yes :)

julienrf:
@jeroentervoorde I pushed a commit to fix the tests. Could you please take into account my other remarks?

julienrf:
OK, I don’t know why but jenkins ignores my last push…

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
jeroentervoorde: I tried changing it to a normal task but it doesn't compile. I assume because ReloadWorkflowTasks.webpackTask returns an Def.Initialize[Task[...]]

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
jeroentervoorde: Changed

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
jeroentervoorde: Your're right the first test isn't really necessary and the second is plain wrong (i was assuming an empty target dir). I removed them both.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
jeroentervoorde: Missed that one.. Updated now

julienrf:
test this please

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/common.webpack.config.js
```diff
@@ -0,0 +1,26 @@
...
+  module: {
+    loaders: [
+     {
+        test: /\.(jpe?g|png|gif|svg)$/i,
+        loaders: [
+          'file?hash=sha512&digest=hex&name=[hash].[ext]',
+          'image-webpack?bypassOnDebug&optimizationLevel=7&interlaced=false'
+        ]
+      }
+    ],
```
julienrf: According to the [CI (line 1170)](https://ci.appveyor.com/project/julienrf/scalajs-bundler/build/217#L1170) it seems that at some point a `require('fs')` is performed *from the browser*… Do you know why? Do the tests run fine on your machine?

jeroentervoorde: I don't know. I ignored it because everything seems to run fine. I'll try to fix it tonight or tomorrow.

julienrf: At least the tests passed on the *nix CI. We have so many troubles with Windows support :(

manual/src/ornate/cookbook.md
```diff
@@ -53,6 +53,63 @@ More fine-grained control over the list of monitored files is possible by overri
...
+## Sharing webpack configuration among configuration files {#shared-config}
+
+In addition to the configured webpack config file, all .js files in the project base directory
+(as configured using the `webpackResources` setting) are copied to the target directory so they can be imported
+from the various configuration files.
+
+Here are the steps to share the loader configuration among your prod and dev config files. This
+uses webpack-merge for convenience the same result could be accomplished using plain js only.
+
+1. Put configuration it in a common.webpack.config.js file:
```
julienrf: Typo: extra “it”

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/common.webpack.config.js
jeroentervoorde: Hi. I fixed it using the additional "node" configuration described here:
https://github.com/pugjs/pug-loader/issues/8

julienrf: Hmm, but that’s weird because the final target is the web browser, not node, right?

jeroentervoorde: I guess so.. Is the target something i can/should change? I couldn't find it in the code.

julienrf:
Thank you so much!

PR #126, opened 2017-04-28
Ability to add custom options to NPM config.

cranst0n:
I found this ability useful in my own workflow and thought it may be useful to others.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
@@ -22,7 +22,7 @@ object JSLike {
 }
 
 /** A convenient wrapper around JS trees */
-final class JS(tree: Tree) extends JSLike(tree) {
+final class JS private(tree: Tree) extends JSLike(tree) {
   import JS.position
   def dot(ident: String): JS = JS(DotSelect(tree, Ident(ident)))
   def bracket(ident: String): JS = JS(BracketSelect(tree, StringLiteral(ident)))
@@ -99,7 +99,7 @@ object JS {
 
 }
 
-final class JSON(tree: Tree) extends JSLike(tree)
+final class JSON private(tree: Tree) extends JSLike(tree)
 
 object JSON {
 
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
@@ -64,10 +64,10 @@ object JS {
     JS(New(varRef("RegExp"), List(StringLiteral(value))))
 
   /** Block of several statements */
-  def block(stats: JSLike*): JS = JS(Block(stats.map(_.tree).to[List]))
+  def block(stats: JS*): JS = JS(Block(stats.map(_.tree).to[List]))
 
   /** Anonymous function definition */
-  def fun(body: JSLike => JSLike): JS = {
+  def fun(body: JS => JSLike): JS = {
     val param = freshIdentifier()
     JS(Function(List(ParamDef(Ident(param), rest = false)), Return(body(ref(param)).tree)))
   }
@@ -91,7 +91,7 @@ object JS {
     )
   }
 
-  def `new`(ctor: JSLike, args: JSLike*): JS = JS(New(ctor.tree, args.map(_.tree).to[List]))
+  def `new`(ctor: JS, args: JSLike*): JS = JS(New(ctor.tree, args.map(_.tree).to[List]))
 
   private val identifierSeq = new AtomicInteger(0)
   private def freshIdentifier(): String =
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
@@ -6,25 +6,39 @@ import org.scalajs.core.ir.Position
 import org.scalajs.core.tools.javascript.Printers
 import org.scalajs.core.tools.javascript.Trees._
 
+private[util] sealed abstract class JSLike(val tree: Tree) {
+  def show: String = tree.show
+  def show(isStat: Boolean = true): String = JSLike.show(tree, isStat)
+  def toJson: String = JSLike.show(tree, isStat = false)
+}
+
+object JSLike {
+  def show(tree: Tree, isStat: Boolean = true): String = {
+    val writer = new java.io.StringWriter
+    val printer = new Printers.JSTreePrinter(writer)
+    printer.printTree(tree, isStat)
+    writer.toString
+  }
+}
+
 /** A convenient wrapper around JS trees */
-final case class JS(tree: Tree) extends AnyVal {
+final class JS(tree: Tree) extends JSLike(tree) {
   import JS.position
   def dot(ident: String): JS = JS(DotSelect(tree, Ident(ident)))
   def bracket(ident: String): JS = JS(BracketSelect(tree, StringLiteral(ident)))
   def bracket(ident: JS): JS = JS(BracketSelect(tree, ident.tree))
   def assign(rhs: JS): JS = JS(Assign(tree, rhs.tree))
   def apply(args: JS*): JS = JS(Apply(tree, args.map(_.tree).to[List]))
-
-  def show: String = tree.show
 }
 
 object JS {
 
   implicit lazy val position: Position = Position.NoPosition
 
-  /** String literal */
-  def str(value: String): JS =
-    JS(StringLiteral(value))
+  def apply(tree: Tree): JS = new JS(tree)
+
+  /** Array literal. */
+  def arr(elems: JS*): JS = JS(ArrayConstr(elems.map(_.tree).to[List]))
 
   /** Boolean literal */
   def bool(value: Boolean): JS = JS(BooleanLiteral(value))
@@ -33,12 +47,12 @@ object JS {
   def obj(fields: (String, JS)*): JS =
     JS(ObjectConstr(fields.map { case (ident, value) => (StringLiteral(ident), value.tree) }.to[List]))
 
+  /** Object literal */
   def objStr(fields: Seq[(String, String)]): JS =
     obj(fields.map { case (k, v) => k -> JS.str(v) }: _*)
 
-  /** Array literal */
-  def arr(elems: JS*): JS =
-    JS(ArrayConstr(elems.map(_.tree).to[List]))
+  /** String literal */
+  def str(value: String): JS = JS(StringLiteral(value))
 
   /** Variable reference */
   def ref(ident: String): JS =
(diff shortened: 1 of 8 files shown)


julienrf:
Hi, thanks for the contribution!

I’m not sure to understand in which situation filling these configuration values is useful, though, could you give more details about your use case?

Also, we should not use the `scalajsbundler.util.JS` type for that because this type can represent any JavaScript tree, whereas a package.json file should only contain a JSON document. (BTW, I’d rather keep the `scalajsbundler.util.JS` thing as a private internal implementation detail ^^)

Or maybe we could break down this `JS` data type into `JSON` and `JS` parts, and use the `JSON` part for this setting…

cranst0n:
So in my own usecase, I use `npmConfig` in conjuction with the `Npm` external command and run [electron-builder](https://github.com/electron-userland/electron-builder) to create an RPM for my application. That requires some additional configuration in `package.json` so I guess maybe I hijacked this plugin a bit to accomplish that goal.

Specifically I added this:
```
npmConfig in Compile := Map(
    "name"        -> str(name.value),
    "version"     -> str(version.value),
    "description" -> str("UI build on electron."),
    "author"      -> str(organizationName.value),
    "license"     -> str("MIT"),
    "main"        -> str("./custom-launcher.js"),
    "homepage"    -> str("http://gitlab.docs.mil/foss-scala/"),
    "scripts" -> obj(
      "start" -> str("electron ./custom-launcher.js"),
      "dist"   -> str("build --linux")
    ),
    "build" -> obj(
      "appId" -> str("foss-positron"),
      "linux" -> obj(
        "packageCategory" -> str("Science"),
        "maintainer"      -> str(organizationName.value),
        "target"          -> arr(str("rpm")),
        "icon"            -> str("build")
      ),
      "win" -> obj(
        "target" -> arr(str("portable"))
      ),
      "portable" -> obj(
        "unicode"      -> bool(false),
        "artifactName" -> str("${productName}-portable-${version}.${ext}")
      )
    )
  )
```

If you think this capability is out of scope I get that. I can always resort to creating some kind of SBT task to get there but figured I would at least share what I ended up with to get to where I needed to go.

julienrf:
Well, I think that the effort to support your use case is fine so I would rather merge this PR. Could you try to apply the following changes, though?
- Rename the key `npmConfig` to `additionalNpmConfig`, to emphasize that part of the configuration (the `dependencies` and `devDependencies` fields) are automatically generated (and you should also mention that point in the key description) ;
- Break down the `JS` type into `JSON` and `JS`, make the `additionalNpmConfig` key of type `JSON` and add more scaladoc to the public API.

cranst0n:
> Break down the JS type into JSON and JS, make the additionalNpmConfig key of type JSON and add more scaladoc to the public API.

I'm not sure how the hierarchy of `JS` and `JSON` would look. Is `JSON` a subset of `JS`? A quick attempt led me to something like:

```scala
sealed abstract class JSLike(val tree: Tree)
final class JS(tree: Tree) extends JSLike(tree) {
final class JSON(tree: Tree) extends JSLike(tree) {
```
But this quickly led me into situations where I'm not sure what the exact relationship would be. As an example, it's not clear to me where a `regex` would land (excerpt from `Webpack`):

```scala
...
JS.obj(
  "test" -> JS.regex("\\.js$"),
  "enforce" -> JS.str("pre"),
  "loader" -> JS.str("source-map-loader")
)
...
```

I may be out of my depth with the semantics here unless you had some simpler plan like just creating a simpler `JSON` type that has no type relationship to `JS`.

julienrf:
Regexps are not `JSON`. You can see [here](http://json.org/) what we should provide constructor for. Note that `package.json` contains a JSON document but `webpack.config.js` contains a JavaScript document (that’s why webpack config files can contain Regexps).

Is it clearer? If that’s still not, I can try to give this idea a try, later :)

cranst0n:
Yes that's clearer. I'll have a quick go at it.

cranst0n:
Let me know if the direction of this is what you had in mind. If not, please feel free to do as you see fit and I can rebase on your implementation.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -79,17 +93,35 @@ object JS {
...
-    writer.toString
-  }
-
   private val identifierSeq = new AtomicInteger(0)
   private def freshIdentifier(): String =
     s"x${identifierSeq.getAndIncrement()}"
 
 }
+
+final class JSON(tree: Tree) extends JSLike(tree)
```
julienrf: That looks perfect excepted that I would put this constructor private so that users have to use the smart constructors defined in tho `JSON` object.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -79,17 +93,35 @@ object JS {
...
+  def apply(tree: Tree): JSON = new JSON(tree)
+
+  /** Array literal. */
+  def arr(elems: JSON*): JSON = JSON(ArrayConstr(elems.map(_.tree).to[List]))
+
+  /** Boolean literal */
+  def bool(value: Boolean): JSON = JSON(BooleanLiteral(value))
+
+  /** Object literal */
+  def obj(fields: (String, JSON)*): JSON =
```
julienrf: Do you think that would work to:
- remove duplicate definitions between `JS` and `JSON` smart constructors (e.g. `str`, `obj`, etc.) ;
- replace references to `JSON` and `JS` with `JSLike`, when needed.

For instance the type signature of `JSON.obj` would be:

~~~ scala
def obj(fields: (String, JSLike)*): JSON
~~~

The type signature of `JS.let` would be:

~~~ scala
def let(value: JSLike)(usage: JS => JSLike): JS
~~~

If you think that would be too much effort I’m happy to merge the PR in its current state (can you just update the scaladoc description to mention that the `dependencies` and `devDependencies` are automatically generated from the corresponding sbt settings?).

cranst0n: That crossed my mind at first but that ends up allowing mixing of `JS` and `JSON` which we wanted to avoid, right?. For example if `JSON.obj` accepts `JSLike` key/values as you suggest, this compiles:

```scala
scala> val x = JSON.obj("a" -> JSON.bool(true), "b" -> JS.regex("[abAZ]"))
x: scalajsbundler.util.JSON = scalajsbundler.util.JSON@1084acc1
```

julienrf:
Oh, indeed I was wrong about JSON.obj: we must stay in the JSON world. But there are places where we require JS values although we should accept JSLike values. 

Anyway, that's not super important :)

cranst0n:
I *think* I've adjusted the required types appropriately but let me know if things don't look right.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -6,39 +6,53 @@ import org.scalajs.core.ir.Position
...
+  /** Array literal. */
+  def arr(elems: JSLike*): JS = JS(ArrayConstr(elems.map(_.tree).to[List]))
 
   /** Boolean literal */
   def bool(value: Boolean): JS = JS(BooleanLiteral(value))
 
   /** Object literal */
-  def obj(fields: (String, JS)*): JS =
+  def obj(fields: (String, JSLike)*): JS =
     JS(ObjectConstr(fields.map { case (ident, value) => (StringLiteral(ident), value.tree) }.to[List]))
```
julienrf: Can we remove this method now instead of duplicating `JSON.obj`?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -6,39 +6,53 @@ import org.scalajs.core.ir.Position
...
 
+  /** Object literal */
   def objStr(fields: Seq[(String, String)]): JS =
     obj(fields.map { case (k, v) => k -> JS.str(v) }: _*)
```
julienrf: Same for this one.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -6,39 +6,53 @@ import org.scalajs.core.ir.Position
...
 
-  /** Array literal */
-  def arr(elems: JS*): JS =
-    JS(ArrayConstr(elems.map(_.tree).to[List]))
+  /** String literal */
+  def str(value: String): JS = JS(StringLiteral(value))
```
julienrf: And this one.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -50,10 +64,10 @@ object JS {
     JS(New(varRef("RegExp"), List(StringLiteral(value))))
 
   /** Block of several statements */
-  def block(stats: JS*): JS = JS(Block(stats.map(_.tree).to[List]))
+  def block(stats: JSLike*): JS = JS(Block(stats.map(_.tree).to[List]))
```
julienrf: I would keep `JS` here because that would not make a lot of sense to allow a (pure) `JSON` value in a list of statements.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -50,10 +64,10 @@ object JS {
...
 
   /** Anonymous function definition */
-  def fun(body: JS => JS): JS = {
+  def fun(body: JSLike => JSLike): JS = {
```
julienrf: I would type the body as follows instead: `JS => JSLike`

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
```diff
@@ -77,19 +91,37 @@ object JS {
...
-
-  def toJson(obj: JS): String = show(obj.tree, isStat = false)
-
-  def show(tree: Tree, isStat: Boolean = true): String = {
-    val writer = new java.io.StringWriter
-    val printer = new Printers.JSTreePrinter(writer)
-    printer.printTree(tree, isStat)
-    writer.toString
-  }
+  def `new`(ctor: JSLike, args: JSLike*): JS = JS(New(ctor.tree, args.map(_.tree).to[List]))
```
julienrf: What about `(ctor: JS, args: JSLike*): JS` ?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/JS.scala
cranst0n: It's not quite duplicated though. The JSON definition takes `(String, JSON)*` to explicitly disallow any `JS` stuff to be mixed in. Removing `JS.obj` will make [this](https://github.com/scalacenter/scalajs-bundler/blob/fb8064bc888f2c829d86daaea6e89144c507d8fe/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala#L76) invalid since it's adding a `RegExp` type.

julienrf: Yes, good point, thanks for making it clearer!

julienrf:
Thanks a lot for your time working on that!

cranst0n:
:+1: Thanks for working with me on this. Really helpful project!

Issue #114, opened 2017-03-22
scalaJSLauncher is deprecated in Scala.js v0.6.15


garhirunrat:
https://github.com/scala-js/scala-js/blob/502bde1b1d6046ea531df95abba64b61265168ab/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala#L255

persistLauncher is deprecated in favor of scalaJSUseMainModuleInitializer
https://www.scala-js.org/news/2017/03/21/announcing-scalajs-0.6.15/

julienrf:
Hi, yes that should work. Can you give more details on how to reproduce your issue (maybe open a different issue?)

benderpremier:
Yes sure, so basically what I did to try out this functionality was:

1. I created a dummy scalajs object:
```scala
package example

import org.scalajs.dom

import scala.scalajs.js
import scala.scalajs.js.annotation.{JSExport, JSExportTopLevel}

@JSExportTopLevel("TicTacToe")
object TicTacToe extends js.JSApp{
  @JSExport
  def main(): Unit = {
    dom.window.alert("Hi from Scala-js-dom")
  }
}

```

Then my sbt file looked like:

```scala
scalaVersion := "2.12.2"
libraryDependencies += "org.scala-js" %%% "scalajs-dom" % "0.9.2",
      
libraryDependencies += "com.github.japgolly.scalajs-react" %%% "core" % "1.0.0",
  
npmDependencies in Compile ++= Seq(
        "react" -> "15.5.4",
        "react-dom" -> "15.5.4"
      )

webpackConfigFile := Some(baseDirectory.value / "custom.webpack.config.js")

import org.scalajs.core.tools.io.{VirtualJSFile, FileVirtualJSFile}

// Use the output of Scala.js as a “launcher”
scalaJSLauncher in (Compile, fullOptJS) := {
  Attributed.blank[VirtualJSFile](FileVirtualJSFile((fullOptJS in Compile).value.data))
}

```
and custom.webpack.config:

```scala

var webpack = require('webpack');

module.exports = require('./scalajs.webpack.config');
module.exports.target = "web";
// Expose the Scala.js artifact as a library in the `sjs` nampespace
module.exports.output.library = 'sjs';
```
The issues I was seeing were:

1. the "Uncaught ReferenceError: exports is not defined"
[...]

PR #131, opened 2017-06-07
Handle ImportWithGlobalFallback from scalajs 0.6.17

mariusmuja:
The reload workflow doesn't work properly in case of dependencies that use scalajs 0.6.17 and use @JSImport with globalFallback. This PR adds support for these in the reload workflow.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ val `sbt-scalajs-bundler` =
       sbtPlugin := true,
       name := "sbt-scalajs-bundler",
       description := "Module bundler for Scala.js projects",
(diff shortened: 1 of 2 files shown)


julienrf:
Hi, thanks for the contribution!

I think we should first publish a new release of the current state of scalajs-bundler and then create a new branch to support Scala.js 0.6.15+. However I think we should probably fix #114 before doing any release targeting Scala.js 0.6.15+.

julienrf:
Hi @mariusmuja, your PR has been rebased in #141 so I’m closing this one. Thanks for your contribution!

Issue #133, opened 2017-06-12
Automatically populate npmAdditionalNpmConfig description/licence/repository 


olafurpg:
We're using scalajs-bundler in scalameta and get this error message
```
[error] npm WARN test No description
[error] npm WARN test No repository field.
[error] npm WARN test No license field.
```
https://travis-ci.org/scalameta/scalameta/jobs/241610752#L508-L510

I believe it's possible to silence those errors with npmAdditionalNpmConfig: https://github.com/scalacenter/scalajs-bundler/blob/ad42d1cdb77ae4b6eafa50b0e1013c1dd090aecc/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala#L140

It would be nice if those required fields would be set automatically, however.

julienrf:
Indeed, the new `additionalNpmConfig` key makes it easy for users to implement this.

But I agree that we could provide it out of the box :)

japgolly:
You can also avoid this by setting `"private": true` in the generated `package.json`.

eg. default:
```
> cat package.json
{
}

> npm install     
npm WARN node No description
npm WARN node No repository field.
npm WARN node No license field.
```
and private
```
> cat package.json
{
  "private": true
}

> npm install     
```

PR #135, opened 2017-06-20
Make package.json private by default (Issue #133)

Freddex:
Sets "private" property in the generated package.json to private, to avoid error messages on build. (See #133 for details)

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
@@ -51,6 +51,7 @@ object PackageJson {
       JSON.obj(
         (
           additionalNpmConfig.toSeq :+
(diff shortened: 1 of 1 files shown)


julienrf:
I’d rather define that as a [default value for `additionalNpmConfig`](https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala#L547):

~~~ scala
additionalNpmConfig := Map("private" -> JSON.bool(true))
~~~

easel:
I find that setting the license is also useful, see https://github.com/scalacenter/scalajs-bundler/pull/149/files#diff-5b0715ef9d50e5870cb4b6350586a5e6R520

easel:
I think we got this one covered in another PR. Thanks!

PR #146, opened 2017-07-16
Global modules with jsdom unit testing

### LINKED ISSUE #145: "How to use an existing facade ..." cookbook recipe doesn't describe testing with jsdom case
I've found that this recipe https://scalacenter.github.io/scalajs-bundler/cookbook.html#global-namespace works well for standard dev or prod build scenario. When it comes to testing with `requiresDOM in Test := true` case it fails to load global modules. 

From what I've found in the source it looks like testing is handled differently: https://github.com/scalacenter/scalajs-bundler/blob/8d1d7fed5ee87e5e526b523540d3b377f90fe1ad/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala#L652

I've managed to get my setup working but with slightly different webpack config:

```js
var globalModules = {
  moment: "moment"
};

var config = {
  module: {
    loaders: [
      {
        // Force require global modules
        test: /.*-fastopt\.js$/,
        loader: "imports-loader?" + Object.keys(globalModules).map(function(modName) {
          return modName + "=" + globalModules[modName];
        }).join(',')
      },
      Object.keys(globalModules).map(function(modName) {
        // Expose global modules
        return {
          test: require.resolve(modName),
          loader: "expose-loader?" + globalModules[modName]
        }
      })
    ]
  }
};

module.exports = config;
```

It uses [imports-loader#0.7.0](https://github.com/webpack-contrib/imports-loader/tree/v0.7.0) to force require of global modules so prepending globals to entries is no longer needed. It can be put into `common.webpack.config.js` and then reused in both `dev.webpack.config.js` and `test.webpack.config.js` by merging with [webpack-merge](https://github.com/survivejs/webpack-merge).

Maybe it's worth updating the cookbook for this scenario? Or is there maybe a easier way to achieve the same resul?

pwlmaciejewski:
Fixes #145.

It adds new test case `global-modules-with-jsdom-unit-testing`
and introduces new webpack config that works for both dev and testing build.

It changes cookbook to reference newly added test case config.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target/
+.idea
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -192,22 +192,22 @@ dependency on the facade and to enable the `ScalaJSBundlerPlugin` plugin.
 
 ## How to use an existing facade assuming the JS library to be exposed to the global namespace? {#global-namespace}
 
-Webpack is able to expose modules to the global namespace by using a custom loader:
-[expose-loader](https://github.com/webpack/expose-loader). Thus, you can write a custom webpack configuration
-file that uses this loader to expose the required modules to the global namespace. Typically, this file
-will look like this:
+Webpack is able to require external modules by using [imports-loader](https://github.com/webpack-contrib/imports-loader) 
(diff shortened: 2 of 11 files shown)


julienrf:
test this please

julienrf:
Thanks @pwlmaciejewski!

PR #161, opened 2017-08-21
Suppress NPM package.json errors

easel:
- Sets the default package.json to private = true, UNLICENSED
  with a blank description. Eliminates the extra error spam from newer
  versions of npm complaining about those fields not being set.


julienrf:
Thanks!

easel:
Thanks for merging!

PR #163, opened 2017-08-21
Add json output support for var and int literals

easel:


sjrd:
This one I would suggest be kept in the PR that actually uses these things. The rationale, compared to the others, it's that this commit alone does not provide any value to the users. It should get into master if and only if another commit follows up with using it, otherwise we introduce dead code. So it does not make sense as an individual PR.

easel:
Ok, that's fine. I've got it in that branch as well and can leave it there. Closing this one out.

PR #160, opened 2017-08-19
Webpack 3.X support.

MasseGuillaume:

PR #166, opened 2017-08-21
Enable Travis CI

julienrf:
This PR cherry picks commits from #160 that are specific to Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache:
     - $HOME/.coursier/cache
 before_install:
   # Install Node.js and yarn.
-  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
+  - nvm install 6
+  - nvm use 6
+  - node --version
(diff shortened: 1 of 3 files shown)


julienrf:
I don’t really care about coursier. Is that really important?

julienrf:
A more important problem for me is that the tests are failing on Travis without clear clues on the reason…

MasseGuillaume:
@julienrf having coursier in scripted will speed up the ci time a lot.

julienrf:
> having coursier in scripted will speed up the ci time a lot.

Why if we cache ivy’s resolutions?

julienrf:
I just tried and didn’t really notice a performance improvement (or maybe 3%). But that’s fine…

sjrd:
With a properly set up cache, the default ivy resolution is just fine.

I only had to resort to force coursier on Travis when I had to fetch and build all of Scala.js locally (on the Travis machine) before testing things that depended on master, because in that case the build would not even *complete* with ivy, therefore never putting things in the cache. When I did, in order not to pollute the repo everywhere with coursier setup, I did it with [these lines](https://github.com/scala-js/jsdependencies/blob/dd9a9c5365d0f85299ca69ff47d15308c4dbdf67/.travis.yml#L11-L14) and [that line](https://github.com/scala-js/jsdependencies/blob/dd9a9c5365d0f85299ca69ff47d15308c4dbdf67/.travis.yml#L43) in `.travis.yml`.

PR #171, opened 2017-08-31
Add all frontend project resource folders to watch scala.js watch list

easel:
- Causes the bundler to fire on changes to plain .js files in src/main/resources


julienrf:
Would it be possible to add any test for this feature?

easel:
Well, we haven't got a pattern for such a thing yet. In general, it would be nice to script out tests for all the "changes should fire a rebuild scenarios" but I'm not entirely sure how to get from here to there at this point.

PR #172, opened 2017-08-31
Use npm install instead of npm update for package bootstrap

easel:
- Npm update behavior has changed in versions > ~5.0.0

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -572,7 +572,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
             if (useYarn.value) {
               Yarn.run("install", "--non-interactive")(targetDir, log)
             } else {
-              Npm.run("update")(targetDir, log)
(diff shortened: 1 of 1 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -572,7 +572,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
             if (useYarn.value) {
               Yarn.run("install", "--non-interactive")(targetDir, log)
             } else {
-              Npm.run("update")(targetDir, log)
+              Npm.run("install", ".")(targetDir, log)
```
julienrf: What’s the benefit of the `.`?

easel: Just explicitness:

easel: That said, I glazed over the first line when I was reading that help the first time, so it's apparently documented behavior to look for package.json in the current folder.

cquiroz:
Should this make it work on npm >= 5.x

easel:
Yes, should resolve it.

PR #175, opened 2017-09-13
Webpack 3.X is now the default version

MasseGuillaume:

PR #176, opened 2017-09-13
Dev server install

MasseGuillaume:
Depends on: https://github.com/scalacenter/scalajs-bundler/pull/175


julienrf:
Can you create a test showing what these changes fix?

MasseGuillaume:
Sure, but it's going to take some time. Here is the test plan:

create a config with css loading + dev-server with hot reloading.
touch the css
obtained: the page does not hot reload the css
expected: the css is hot reloaded

ngthanhtrung:
I also encountered this and spent a huge time debugging it. After figuring out the problem I also found out that it has been fixed. Kudos to @MasseGuillaume!

PR #177, opened 2017-09-14
Scope bundling mode per scalajs stage

MasseGuillaume:


MasseGuillaume:
You can set

```
webpackBundlingMode := BundlingMode.Library()
```

to set a value for both stages

or

```scala
webpackBundlingMode in fastOptJS := BundlingMode.LibraryOnly(),
webpackBundlingMode in fullOptJS := BundlingMode.Application,
```

or individually

julienrf:
Thanks :)

Issue #150, opened 2017-08-13
Publish for SBT 1.0.0


shadaj:
Now that SBT 1.0.0 has been released, it would be great to have the plugin support it.

aparo:
Is there any update to this issue?

julienrf:
I didn't work on it but any contribution is welcome. I guess the issue is ready to be solved because the Scala.js plugin already supports sbt 1.0. We should also check for sbt-web-scalajs.

nafg:
sbt-web-scalajs was published for sbt 1. https://github.com/vmunier/sbt-web-scalajs/pull/38#issuecomment-325633686

rleibman:
I've got a good chunk of it done, I think. I had to upgrade a bunch of the plugins and modify the build.sbt slightly. The only thing currently blocking me is that there's no version of the OrnatePlugin for sbt 1.0,

I'm working holistically (i.e. not understanding anything) at this point, I just want things to compile.

sjrd:
I highly recommend not to try and update the *build* itself to sbt 1.0. This requires that all the sbt plugins used by *the build* (as opposed to by the plugin) be available for 1.0.

Instead, the build should use sbt 0.13.16, and then you can use `^^1.0.1` to build the plugin itself against sbt 1.0. This is by a large margin the path of least resistance.

rleibman:
I'm not sure what you mean by ^^1.0.1, where do I type that? is that something I add to the build.sbt, or is it a command in sbt?

julienrf:
I agree with @sjrd, migrating the plugin only (without migrating the build) is quite easy actually (we just have to change some file manipulation and `sbt.Process` usage). Actually I tried to do the migration a few days ago, it works OK but I got a dozen of false positive errors because of this issue: https://github.com/sbt/sbt/issues/3299. I was thinking of waiting for it to be fixed before actually making the migration.

jonas:
Started work on this (sorry @rleibman, didn't see your message until now).

#179 bumps dependencies and fixes several problems when building with sbt 1.0 such as
```
[error] Problem: `useYarn` is inside the if expression of a regular task.
[error]   Regular tasks always evaluate task inside the bodies of if expressions.
```

[...]

PR #179, opened 2017-09-15
Cross build for sbt 1.0

jonas:
Bumps dependencies and fixes warnings when compiling with sbt 1.0.

References #150


build.sbt
```diff
@@ -88,9 +87,7 @@ val `scalajs-bundler` =
         commitNextVersion,
         pushChanges,
         releaseStepTask(GhPagesKeys.pushSite in manual)
-      ),
-      scriptedLaunchOpts += "-Dplugin.version=" + version.value,
-      scriptedBufferLog := false
```
jonas: Not sure why the top-level project contains settings for the scripted plugins, so I've removed them.

jonas:
I'm seeing a lot of library dependencies warnings, should the sbt-scalajs version be updated in the scripted tests
```
info] [warn] There may be incompatibilities among your library dependencies.
[info] [warn] Here are some of the libraries that were evicted:
[info] [warn] 	* org.scala-js:sbt-scalajs:0.6.18 -> 0.6.19
[info] [warn] Run 'evicted' to see detailed eviction warnings
```

sjrd:
See https://contributors.scala-lang.org/t/a-guide-on-binary-compatibility-need-your-input/1089/19?u=sjrd

jonas:
@sjrd Not sure I understand. Are you saying that we should the sbt-scalajs plugin version as is?

sjrd:
Yes, basically. Or at least that the reporting warning is not a concern and should not itself be the cause to upgrade.

jonas:
Currently working on getting the scripted tests to run for sbt 1.0. I'll need some guidance from the maintainers on how to approach this. Some tests only support sbt 0.13 (for example those depending on sbt-play), other tests we want to run for sbt 1.0 needs to be upgraded to use Scala.js 0.6.19. In addition, we need to remove the `project/build.properties` files from scripted tests that should run for sbt 1.0.

build.sbt
```diff
@@ -135,3 +151,17 @@ lazy val noPublishSettings =
...
+def filterScripted(excludeF: PartialFunction[String, Seq[File => PathFinder]]) = Def.taskDyn {
+  val base = sbtTestDirectory.value
+  val allTests = base * AllPassFilter * AllPassFilter filter { _.isDirectory }
+  val exclude = excludeF((sbtBinaryVersion in pluginCrossBuild).value)
+  val tests = exclude.map(_.apply(base)).foldLeft(allTests)(
+    (remaining, excluded) => remaining --- excluded).get
+  def args = tests.map(Path.relativeTo(base)).flatten.mkString(" ")
+
+  if (tests.nonEmpty)
+    Def.task(scripted.toTask(" " + args).value)
```
jonas: This is a super generic way to filter scripted tests based on @xuwei-k's amazing work on https://github.com/sbt/sbt-site/pull/107. If necessary we can simplify it to filter only the last directory, e.g. under `sbt/sbt-{,web}-scalajs-bundle` and include some defaults such as include tests matching `*sbt-${sbtVersion}*`. Let me know what you think/

sbt-scalajs-bundler/src/main/scala-sbt-0.13/scalajsbundler/Compat.scala
```diff
@@ -0,0 +1,12 @@
...
+object Compat {
+  type Process = sbt.Process
+  val Process = sbt.Process
+  type ProcessLogger = sbt.ProcessLogger
+
+  // Mimics scala.sys.process.ProcessLogger
+  implicit class ProcessLoggerOps(val self: ProcessLogger) extends AnyVal {
+    def out(msg: String) = self.info(msg)
+  }
+}
```
julienrf: Why not use `scala.sys.Process` everywhere instead of introducing this abstraction over `scala.sys.Process` and `sbt.Process`?

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/additonalNpmConfig/project/build.properties
```diff
@@ -1 +0,0 @@
-sbt.version=0.13.15
```
julienrf: Why remove that? I think keeping it (and setting it to `1.0.1`) makes it possible to load the project whatever sbt version is installed.

sbt-scalajs-bundler/src/main/scala-sbt-0.13/scalajsbundler/Compat.scala
jonas: I will give it a try.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/additonalNpmConfig/project/build.properties
jonas: If we keep it, it will force the test to only work with that specific sbt version. Most tests I assume we want to run for both versions of sbt. Those we want to fix to a specific sbt version will have to be configured so the CI can filter them out. I've already added a mechanism to do this as pointed out elsewhere.

build.sbt
jonas: Based on @julienrf 's comment we can make this filtering based on the content of `project/build.properties`. If the file is missing, run for all sbt versions else parse the binary version and run it if it matches `sbtBinaryVersion in pluginCrossBuild`.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -492,10 +492,11 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
       val installDir = target.value / "scalajs-bundler-jsdom"
       val log = streams.value.log
       val jsdomVersion = (version in installJsdom).value
+      val usingYarn = useYarn.value
       if (!installDir.exists()) {
         log.info(s"Installing jsdom in ${installDir.absolutePath}")
         IO.createDirectory(installDir)
-        install(installDir, useYarn.value, log)(s"jsdom@$jsdomVersion")
+        install(installDir, usingYarn, log)(s"jsdom@$jsdomVersion")
```
julienrf: I wish https://github.com/sbt/sbt/issues/3299 could be fixed soon so that we don’t have to introduce these redundant identifiers (ie `usingYarn` instead of `useYarn.value`)

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/additonalNpmConfig/project/build.properties
julienrf: OK I see! I didn’t notice that the tests were executed with both sbt 1.0 and 0.13. This is great actually :) We should not have to introduce a mechanism testing whether a test should be run with sbt 1.0 or 0.13 only.

sbt-scalajs-bundler/src/main/scala-sbt-0.13/scalajsbundler/Compat.scala
jonas: Done in 8f2e318

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
jonas: 4b2d2f9 updates to newly released sbt 1.0.2 and allows to remove these workarounds.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/additonalNpmConfig/project/build.properties
jonas: We need that mechanism for at least `sbt-web-scalajs-bundler/play` until the Play sbt plugin has been published for sbt 1.0 (https://github.com/playframework/playframework/issues/7261). However, based on @sjrd's concern, we should probably also add tests that target the minimum version of sbt 0.13 and sbt 1.0 that scalajs-bundler supports to ensure documentation about those requirements stays up to date.

build.sbt
julienrf: If I remember correctly that was needed to run the `scripted` command from the root project. I noticed that you use the `runScripted` now instead. What’s the difference between them?

build.sbt
```diff
@@ -135,3 +144,30 @@ lazy val noPublishSettings =
     publish := (),
     publishLocal := ()
   )
+
+def runScriptedTask = Def.taskDyn {
+  val sbtBinVersion = (sbtBinaryVersion in pluginCrossBuild).value
+  val base = sbtTestDirectory.value
+  val testDirectoryFinder = base * AllPassFilter * AllPassFilter filter { _.isDirectory }
```
julienrf: Why does that find test directories?

jonas: It works similar to the shell glob `.../sbt-test/*/*`

jonas:
OK, squashed and added tests for minimum versions of sbt.

build.sbt
jonas: Completely missed this comment. I will update CONTRIBUTNG.md to mention the new `runScripted` task. A future improvement could be to run the supported scripted tests as part of the normal test target

```
test in Test := Def.sequential(
   test in Test,
   runScriptedTask
).value,
```

appveyor.yml
```diff
@@ -7,8 +7,8 @@ install:
 build_script:
   - sbt clean compile
 test_script:
-  - sbt test scripted
+  - sbt test runScripted
```
julienrf: Do you know if it’s possible to have the same matrix thing as we have with travis?

build.sbt
jonas: Updated CONTRIBUTING.md to mention `runScripted`.

appveyor.yml
jonas: Done in the latest version. Take note of the comment about special handling of `^`. It looks related with https://github.com/appveyor/ci/issues/1500.

julienrf:
This is great work, thanks a lot Jonas!

PR #183, opened 2017-09-29
Cookbook: resolving global namespace conflict (fixes #178).

### LINKED ISSUE #178: Add to docs, that BundlerMode.Library* may be not compatible with some js libs via global scope in production
Hi. 

This crazy issue occurs only in `fullOptJS`.

We have a really FAT internal project, that uses `leaflet.js` from scala.js.
Historically, Leaflet is living in global `L` variable, and only last months there are some support for CommonJS appearing.

But Leaflet still binds global `window.L` for backward compatibility (btw, i'm not tested latest version of leaflet).

After switching to `scalajs-bundler-0.8.0`, project stopped to work in production `fullOptJS` mode, when `webpackBundlingMode := BundlingMode.LibraryOnly()` or `LibraryAndApplication()` also.
```
TypeError: L is not a function
UY(http://localhost:9000/vassets/d1a0d84e7b66ee8617dd736af8cca5ba-app-opt.js:2032:130)
... (tl;dr useless stacktrace) ...
```

As you guessed it, leaflet global `L` somehow conflicts with `var L=` from minified `app-sjs-opt.js` source.

!!!! But when `webpackBundlingMode` is defaulted, **production build works as expected**.

I unsure, it is a bug or not.
Possible, documentation highlights or warnings will be enought...

UPD1: Also, i cannot provide test-case, because the issue is very floating and unstable.

UPD2: Leaflet 1.2.0 (latest version) -- same problem. Used #177 as workaround.

helllamer:
Add cookbook recipe about resolving problems like described in #178 or in more complex use-cases.

### DIFF
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -280,3 +280,80 @@ reload on every change:
 ~~~
 webpackDevServerExtraArgs := Seq("--inline")
 ~~~
+
+## How to resolve conflicts in global namespace between webpack output and 3rd-party js libs?
+
+The only reliable way is to isolate such js libraries from global scope.
+
+Many js libs modify global scope, register global vars, etc. Let's look on these two:
+~~~ scala
+npmDependencies in Compile ++= Seq(
+  "leaflet" -> "1.2.0",
+  "leaflet.markercluster" -> "1.1.0"
+)
+~~~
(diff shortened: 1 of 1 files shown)


helllamer:
AppVeyor build failing due to very internal problems, unrelated to this or any other commits.

Any comments on this PR?

PR #184, opened 2017-10-09
Publish from Travis-CI

julienrf:
The goal of this PR is to simplify the release process: any contributor that has write access to this repository can draft and publish a release.


CONTRIBUTING.md
```diff
@@ -65,6 +69,9 @@ $ sbt publishLocal
 $ sbt manual/previewSite
 ~~~
 
-## General recommendations
+## Publish a release
 
-Test and document each added feature.
+- Check that the `changelog.md` file is up to date (in the `manual` project)
+- Draft a new release on [GitHub](https://github.com/scalacenter/scalajs-bundler/releases/new)
```
jvican: Why is this necessary? When we push a tag from git, GitHub automatically creates a GitHub release with the tag number, and compresses the repository in a zip that you can download.

build.sbt
```diff
@@ -66,77 +67,47 @@ val manual =
...
-      }
-    },
-    // fixed in https://github.com/sbt/sbt/pull/3397 (for sbt 0.13.17)
-    sbtBinaryVersion in update := (sbtBinaryVersion in pluginCrossBuild).value,
-    runScripted := runScriptedTask.value,
-    scriptedLaunchOpts += "-Dplugin.version=" + version.value,
-    scriptedBufferLog := false
-  )
+inThisBuild(List(
+  pgpPublicRing := file("./travis/local.pubring.asc"),
```
jvican: This needs to be scoped in `Global`.

build.sbt
```diff
@@ -66,77 +67,47 @@ val manual =
...
+  pgpSecretRing := file("./travis/local.secring.asc"),
+  releaseEarlyWith := SonatypePublisher,
+  scalacOptions ++= Seq(
```
jvican: Are these being used at all? I think they're ignored by all the projects.

build.sbt
jvican: In the sbt-release-early docs, I show how you can scope it in global with a similar construct as `inThisBuild`.

CONTRIBUTING.md
julienrf: You’re right either way works (push a git tag or publish a tag from GitHub). Since we are not currently using GitHub releases to show our changelog we don’t really care about manually creating a GitHub release, so I should rather mention the way that consists in pushing a tag.

build.sbt
julienrf: Why are they ignored? My goal is that both `sbt-scalajs-bundler` and `sbt-web-scalajs-bundler` use these options.

julienrf:
Thanks Jorge for the review! I’ll address your comments soon.

CONTRIBUTING.md
jvican: Yeah, just say "push a tag" with github (and maybe inline the command to do so). The CI will take care of everything.

build.sbt
jvican: They're ignored because `scalacOptions` is defined in the configurations of the projects, not the projects. Therefore, sbt will never delegate to this one. This is confirmed by the "Provided by" section:
  
```scala
sbt @ caca > inspect scalacOptions
[info] Task: scala.collection.Seq[java.lang.String]
[info] Description:
[info] 	Options for the Scala compiler.
[info] Provided by:
[info] 	{file:/data/rw/code/scala/caca/}caca/compile:scalacOptions
[info] Defined at:
[info] 	(sbt.Classpaths) Defaults.scala:1871
[info] 	(sbt.Defaults) Defaults.scala:1041
[info] Dependencies:
[info] 	Compile/sbtPlugin
[info] 	Compile/settingsData
[info] 	Compile/buildDependencies
[info] 	Compile/update
[info] 	Compile/autoCompilerPlugins
[info] 	Compile/thisProjectRef
[info] Reverse dependencies:
[info] 	Compile/ensimeCompileOnly/scalacOptions
[info] 	ensimeScalacOptions
[info] 	Compile/argsFileContents
[info] Delegates:
[info] 	Compile/scalacOptions
[info] 	scalacOptions
[info] 	ThisBuild/Compile/scalacOptions
[info] 	ThisBuild/scalacOptions
[info] 	Zero/Compile/scalacOptions
[info] 	Global/scalacOptions
[info] Related:
[info] 	Test/ensimeCompileOnly/scalacOptions
[info] 	Compile/ensimeCompileOnly/scalacOptions
[info] 	Global/scalacOptions
[info] 	Test/scalacOptions
```

sjrd: That's not true. Because the original definition of  `scalacOptions in (proj, Compile)` is copied from `scalacOptions in proj`, which itself is copied from `scalacOptions in ThisBuild`.

Those settings are not ignored.

jvican: Yes, you're right.

fdietze:
While you're at the travis configuration: I think the build times could be improved a bit with more travis caching:
```yml
cache:
  yarn: true
  directories:
  - $HOME/.ivy2/cache
  - $HOME/.ivy2/local
  - $HOME/.sbt/boot/
```

julienrf:
@fdietze Currently we cache `$HOME/.sbt`, so it includes `$HOME/.sbt/boot`, right? Also, what would be the benefit of caching `$HOME/.ivy2/local`? We always do fresh `publishLocal` when we run the scripted tests so anyway the contents of this repository is going to be overwritten, right?

fdietze:
@julienrf I think you are right. I overlooked `$HOME/.sbt` in the diff. And for local, I always have this in my travis conf, just in case I need to `publishLocal` a library to build. But in this case, I think you are right again, because it could even cache too much unnecessary stuff. At least `yarn: true` was an improvement... :sweat_smile:

PR #198, opened 2018-01-09
Fixed #197: Provide Node.js interface through global object in JSDOMNodeJSEnv

### LINKED ISSUE #197: Provide Node.js interface through `global` object in `JSDOMNodeJSEnv`
Currently, in `JSDOMNodeJSEnv` (which is the case when `requiresDOM in Test := true`) there is no Node.js `global` object exposed.

Though, in plain `NodeJSEnv` it is available and can be used as a way to detect/use Node.js environment described [here](https://www.scala-js.org/doc/interoperability/global-scope.html).

As a result there is no way for test frameworks/plugins (like `scoverage`, for example) to access/use Node internal/system modules (`require("fs")` and `require("path")`, for example).

Motivation:
Fixing this issue would be a step forward to make `scoverage` work with `scalajs-bundler` together with corresponding [fix](https://github.com/scoverage/scalac-scoverage-plugin/pull/212) for [this issue in the scoverage-plugin](https://github.com/scoverage/scalac-scoverage-plugin/issues/183)

viktor-podzigun:
Fix for issue #197

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/JSDOMNodeJSEnv.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/JSDOMNodeJSEnv.scala
@@ -94,6 +94,7 @@ class JSDOMNodeJSEnv(
            |    var window = dom.window;
            |    window["__ScalaJSEnv"] = __ScalaJSEnv;
            |    window["scalajsCom"] = global.scalajsCom;
+           |    window["global"] = global;
            |
            |    var scriptsSrcs = $scriptsURIsJSArray;
            |    for (var i = 0; i < scriptsSrcs.length; i++) {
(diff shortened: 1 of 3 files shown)


sjrd:
Erm, `JSDOMNodeJSEnv` is not supposed to do that. It is meant to provide a browser-like environment. It happens to be built on top of Node.js, but it must look like a browser. Since a browser doesn't have `global`, `JSDOMNodeJSEnv` shouldn't provide it either.

viktor-podzigun:
Ok, could we then discuss/propose the solution on the issue #197 thread?

I don't mind to discard/close this PR if its a wrong direction.

But I would like to really get it work with `scoverage` and can contribute from my side.

viktor-podzigun:
Closing in favor of https://github.com/scoverage/scalac-scoverage-plugin/pull/456

Issue #85, opened 2017-01-06
Another plugin for npm dependencies only


mathieuleclaire:
For some projects, it is usefull to only fetch the ```JS```s (with ```npmUpdate```) without webpacking all the ```JS```s into one big js. Creating a plugin allowing js fetching without interfering in the ```full/fastOptJS``` build would be very great.

PR #202, opened 2018-01-22
Implements new plugin based on npmUpdate

mathieuleclaire:
Here is an implementation of the plugin disscused in #85. The idea is to use npmUpdate to fetch js files without using webpack. 

I introduced a new case class to define which files are required in the npm archives. I like the fact that needed js are defined in the build.sbt and not in @JSImport("module/foo/bar"). And actually without webpack, I think I have no other option. 

It is called NpmDepsPlugin (not very sexy, it can be updated :)). Typically in a project it is used this way:

```scala
npmDeps in Compile += Dep("ace-builds", "1.2.9", List("ace.js", "mode-scala.js", "theme-github.js"))
```
Then a ```dependencyFile``` key enables to get the file containing all the npm js libs are merged.
That's it !

Typically, you can have this facade in a project:
```scala
lazy val ace = project.in(file("ace")) enablePlugins (NpmDepsPlugin) settings (defaultSettings) settings (
  scalaJsDom,
  npmDeps in Compile += Dep("ace-builds", aceVersion, List("ace.js", "mode-scala.js", "theme-github.js"))
  )
```
And then:
```scala
lazy val demo = project.in(file("demo")) enablePlugins (NpmDepsPlugin) settings (defaultSettings) settings(
  libraryDependencies += "com.lihaoyi" %%% "scalarx" % rxVersion,
  libraryDependencies += "com.lihaoyi" %%% "sourcecode" % sourceCodeVersion,
  libraryDependencies += "com.github.karasiq" %%% "scalajs-marked" % "1.0.2",
  runDemo := {

    val demoTarget = target.value
    val demoResource = (resourceDirectory in Compile).value

    val jsTarget = demoTarget / "js"
    val demoJS = (fullOptJS in Compile).value

    IO.copyFile(demoJS.data, jsTarget / "demo.js")
    IO.copyFile(dependencyFile.value, jsTarget / "deps.js")

    IO.copyFile(demoResource / "bootstrap-native.html", demoTarget / "index.html")
  }
) dependsOn(ace)
```
Of cours, new npmDeps can be added in the demo project.

It also work if ace is published as a library.

Some points may be upgraded:
- [x] npmUpdate  have been copy / pasted / modified. Peharps, something more generic could be done to have only one implementation to both ```ScalaJSPlugin``` and ```NpmDepsPlugin```
- [x] Keep only ```npmDependencies``` / ```npmUpdate``` stuff in the ```NpmDepsPlugin``` and start a new plugin project inheriting from ```NpmDepsPlugin```  and implementing the jsDependencies part.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
@@ -1,0 +1,55 @@
+package scalajsbundler.sbtplugin
+
+import org.scalajs.core.tools.io.{FileVirtualJSFile, RelativeVirtualFile, VirtualJSFile}
+import org.scalajs.sbtplugin.ScalaJSPlugin.AutoImport.scalaJSNativeLibraries
+import sbt.Keys.{crossTarget, streams}
+import sbt.{Attributed, Def, File, FileFunction, FilesInfo, IO, Task, TaskKey}
+
+import scalajsbundler.{Npm, Yarn}
+import sbt._
+
+import scalajsbundler.BundlerFile.PackageJson
+
+object NpmUpdateTasks {
+  /**
+    * Runs the Npm or Yarn
+    * @param targetDir npm Directory
+    * @param packageJsonFile Json file containing NPM dependencies
+    * @param useYarn Whether to use yarn or npm	
+    * @param jsResources A sequence of javascript resources
+    * @param stream A sbt TaskStream
+    * @return The written npm directory
+    */
+  def npmUpdate(targetDir: File,
+                packageJsonFile: File,
+                useYarn: Boolean,
+                jsResources: Seq[VirtualJSFile with RelativeVirtualFile],
+                streams: Keys.TaskStreams
+                ) = {
+    val log = streams.log
+
+    val cachedActionFunction =
+      FileFunction.cached(
+        streams.cacheDirectory / "scalajsbundler-npm-update",
+        inStyle = FilesInfo.hash
+      ) { _ =>
+        log.info("Updating NPM dependencies")
+        if (useYarn) {
+          Yarn.run("install", "--non-interactive")(targetDir, log)
+        } else {
+          Npm.run("install")(targetDir, log)
+        }
+        jsResources.foreach { resource =>
+          IO.write(targetDir / resource.relativePath, resource.content)
+        }
+        Set.empty
+      }
+
+    cachedActionFunction(Set(packageJsonFile) ++
+      jsResources.collect { case f: FileVirtualJSFile =>
+        f.file
+      }.to[Set])
+
+    targetDir
+  }
+}
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -525,34 +525,12 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
         "license" -> JSON.str("UNLICENSED")
       ),
 
-      npmUpdate := {
-        val log = streams.value.log
-        val targetDir = (crossTarget in npmUpdate).value
-        val jsResources = scalaJSNativeLibraries.value.data
-        val packageJsonFile = scalaJSBundlerPackageJson.value
-
-        val cachedActionFunction =
-          FileFunction.cached(
-            streams.value.cacheDirectory / "scalajsbundler-npm-update",
-            inStyle = FilesInfo.hash
-          ) { _ =>
-            log.info("Updating NPM dependencies")
-            if (useYarn.value) {
-              Yarn.run("install", "--non-interactive")(targetDir, log)
-            } else {
-              Npm.run("install")(targetDir, log)
-            }
-            jsResources.foreach { resource =>
-              IO.write(targetDir / resource.relativePath, resource.content)
-            }
-            Set.empty
-          }
-
-        cachedActionFunction(Set(packageJsonFile.file) ++
-          jsResources.collect { case f: FileVirtualJSFile => f.file }.to[Set])
-
-        targetDir
-      },
+      npmUpdate := NpmUpdateTasks.npmUpdate(
+	(crossTarget in npmUpdate).value,
+	scalaJSBundlerPackageJson.value.file,
+	useYarn.value,
+	scalaJSNativeLibraries.value.data,
+	streams.value),
 
       scalaJSBundlerPackageJson :=
         PackageJsonTasks.writePackageJson(
@@ -802,3 +780,4 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
     }
 
 }
+
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -47,7 +47,7 @@ import scalajsbundler.util.JSON
   * Version of webpack-dev-server to use.
   *
   * {{{
-  *   version in startWebpackDevServer := "2.7.1"
+  *   version in startWebpackDevServer := "2.11.1"
   * }}}
   *
   * == `crossTarget in npmUpdate` ==
@@ -454,7 +454,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
 
     version in webpack := "3.5.5",
 
-    version in startWebpackDevServer := "2.7.1",
+    version in startWebpackDevServer := "2.11.1",
 
     version in installJsdom := "9.9.0",
 
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -325,6 +325,18 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
       taskKey[Seq[File]]("Files that trigger webpack launch")
 
     /**
+      * Additional arguments to webpack
+      *
+      * Defaults to an empty list.
+      *
+      * @group settings
+      */
+    val webpackExtraArgs = SettingKey[Seq[String]](
+      "webpackExtraArgs",
+      "Custom arguments to webpack"
+    )
+
+    /**
       * Whether to use [[https://yarnpkg.com/ Yarn]] to fetch dependencies instead
       * of `npm`. Yarn has a caching mechanism that makes the process faster.
       *
@@ -467,6 +479,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
     // API user can modify it just once.
     webpackMonitoredDirectories := Seq(),
     (includeFilter in webpackMonitoredFiles) := AllPassFilter,
+    webpackExtraArgs := Seq(),
 
     // The defaults are specified at top level.
     webpackDevServerPort := 8080,
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -485,9 +485,10 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
 
     installJsdom := {
       val installDir = target.value / "scalajs-bundler-jsdom"
+      val jsdomDir = installDir / "node_modules" / "jsdom"
       val log = streams.value.log
       val jsdomVersion = (version in installJsdom).value
-      if (!installDir.exists()) {
+      if (!jsdomDir.exists()) {
         log.info(s"Installing jsdom in ${installDir.absolutePath}")
         IO.createDirectory(installDir)
         install(installDir, useYarn.value, log)(s"jsdom@$jsdomVersion")
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -280,3 +280,21 @@ reload on every change:
 ~~~
 webpackDevServerExtraArgs := Seq("--inline")
 ~~~
+
+## How to pass extra parameters to webpack
+
+`scalajs-bundler` invokes `webpack` with a configuration generated either automatically from the build or set with `webpackConfigFile`.
+`webpack` is then called with the following arguments:
+
+~~~
+--bail --config <configfile>
+~~~
+
+You can add extra params to the `webpack` call, for example, to increase debugging
+
+~~~
+webpackExtraArgs := Seq("--profile", "--progress", "true")
+~~~
+
+**Note** Params are passed verbatim, they are not sanitized and could produce errors when passed to webpack.
+In particular, don't attempt to override the `--config` param.
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
@@ -44,7 +44,7 @@ object ExternalCommand {
     */
   def install(installDir: File, useYarn: Boolean, logger: Logger)(npmPackages: String*): Unit =
     if (useYarn) {
-      Yarn.run("add" +: "--non-interactive" +: npmPackages: _*)(installDir, logger)
+      Yarn.run("add" +: "--mutex" +: "network" +: "--non-interactive" +: npmPackages: _*)(installDir, logger)
     } else {
       Npm.run("install" +: npmPackages: _*)(installDir, logger)
     }
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
@@ -75,7 +75,7 @@ object PackageJson {
   /**
     * Resolves multiple occurrences of a dependency to a same package.
     *
-    *  - If all the occurrences refer to the same version, pick this one ;
+    *  - If all the occurrences refer to the same version, pick this one ;
     *  - If they refer to different versions, pick the one defined in `resolutions` (or fail
     *    if there is no such resolution).
     *
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
@@ -136,6 +136,7 @@ object Webpack {
      webpackResources: Seq[File],
      entry: BundlerFile.Application,
      targetDir: File,
+     extraArgs: Seq[String],
      log: Logger
   ): BundlerFile.ApplicationBundle = {
     writeConfigFile(emitSourceMaps, entry, generatedWebpackConfigFile, None, log)
@@ -145,7 +146,8 @@ object Webpack {
       .getOrElse(generatedWebpackConfigFile.file)
 
     log.info("Bundling the application with its NPM dependencies")
-    Webpack.run("--config", configFile.absolutePath)(targetDir, log)
+    val args = extraArgs ++: Seq("--config", configFile.absolutePath)
+    Webpack.run(args: _*)(targetDir, log)
 
     // TODO Support custom webpack config file (the output may be overridden by users)
     val bundle = generatedWebpackConfigFile.asApplicationBundle
@@ -172,6 +174,7 @@ object Webpack {
     webpackResources: Seq[File],
     entryPointFile: BundlerFile.EntryPoint,
     libraryModuleName: String,
+    extraArgs: Seq[String],
     log: Logger
   ): BundlerFile.Library = {
     writeConfigFile(
@@ -186,7 +189,8 @@ object Webpack {
       .map(Webpack.copyCustomWebpackConfigFiles(generatedWebpackConfigFile.targetDir, webpackResources))
       .getOrElse(generatedWebpackConfigFile.file)
 
-    Webpack.run("--config", configFile.absolutePath)(generatedWebpackConfigFile.targetDir, log)
+    val args = extraArgs ++: Seq("--config", configFile.absolutePath)
+    Webpack.run(args: _*)(generatedWebpackConfigFile.targetDir, log)
 
     val library = generatedWebpackConfigFile.asLibrary
     assert(library.file.exists, "Webpack failed to create library file")
@@ -202,7 +206,7 @@ object Webpack {
     */
   def run(args: String*)(workingDir: File, log: Logger): Unit = {
     val webpackBin = workingDir / "node_modules" / "webpack" / "bin" / "webpack"
-    val cmd = "node" +: webpackBin.absolutePath +: "--bail" +: args
+    val cmd = Seq("node", webpackBin.absolutePath, "--bail") ++ args
     Commands.run(cmd, workingDir, log)
     ()
   }
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/LibraryTasks.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/LibraryTasks.scala
@@ -67,7 +67,6 @@ object LibraryTasks {
       assert(ensureModuleKindIsCommonJSModule.value)
       val log = streams.value.log
       val emitSourceMaps = (webpackEmitSourceMaps in stage).value
-      val targetDir = npmUpdate.value
       val customWebpackConfigFile = (webpackConfigFile in stage).value
       val generatedWebpackConfigFile =
         (scalaJSBundlerWebpackConfig in stage).value
@@ -78,6 +77,7 @@ object LibraryTasks {
         Seq(generatedWebpackConfigFile.file, entryPointFile.file) ++
         webpackResourceFiles ++ compileResources
       val cacheLocation = streams.value.cacheDirectory / s"${stage.key.label}-webpack-libraries"
+      val extraArgs = (webpackExtraArgs in stage).value
 
       val cachedActionFunction =
         FileFunction.cached(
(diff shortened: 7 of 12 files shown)


sjrd:
Why doesn't `ScalaJSBundlerPlugin` depend on the new `NpmDepsPlugin`? I would hope that `NpmDepsPlugin` can factor out the npm-related stuff from `ScalaJSBundlerPlugin`, not just be completely independent of it.

sjrd:
In the same spirit, why add a distinct `npmDeps` instead of reusing `npmDependencies`? In any case, having both is super confusing.

mathieuleclaire:
I will try to depends on ScalaJSBundlerPlugin, since I have something working well. It is what I tried first but it rendered things too complex at this stage.

I agree that npmDeps is a new key, but I am not sure I can use @JImport instructions to fetch several js in the same module. In my example, I get 3 files in the ace-builds module. This way, it is easy to define which js you want when you define your build.

sjrd:
> In my example, I get 3 files in the ace-builds module. This way, it is easy to define which js you want when you define your build.

This concern should be separated from the `npmDependencies`. This is a way to tell `resolvedJSDependencies` to take specific modules from `node_modules/` and *execute* them in the context of `run` and `test`. This is very specific. You should have a distinct taskKey for that, that only affects `resolvedJSDependencies`, while `npmDependencies` only affects what gets *installed* (and not necessarily *executed*).

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
```diff
@@ -0,0 +1,152 @@
...
+      }
+
+      val resolvedDependencies = jss.map { f =>
+        ResolvedJSDependency.minimal(FileVirtualJSFile(f))
+      }
+
+      prev.map(_ ++ resolvedDependencies)
+    },
+
+    dependencyFile := (packageMinifiedJSDependencies in Compile).value,
```
sjrd: This is defined, but never *read*. Looks like dead code.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
```diff
@@ -0,0 +1,152 @@
...
+  ) ++ inConfig(Compile)(perConfigSettings)
+
+
+  private lazy val perConfigSettings: Seq[Def.Setting[_]] = Seq(
+    allNpmDeps := NpmDeps.collectFromClasspath((fullClasspath in Compile).value),
+
+    npmUpdate := {
+      val log = streams.value.log
+      val targetDir = (crossTarget in npmUpdate).value
+      val jsResources = scalaJSNativeLibraries.value.data
```
sjrd: Don't introduce new code depending on `scalaJSNativeLibraries`. It's gone in Scala.js 1.x.

mathieuleclaire:
so something like: 
```scala 
npmDependencies += "ace-builds"->  "1.2.9"
otherKey += "ace-builds"-> List("ace.js", "mode-scala.js", "theme-github.js")
``` 
? It's twice as many lines to write

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
mathieuleclaire: dependencyFile is the Key to get the packagedMinifiedJSDependencies. I use it in my demo to build deps.js.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
mathieuleclaire: It is copy/pasted from ScalaJSBundlerPlugin. How to replace this ?

sjrd:
Well, if you really want, you can encode everything in `otherKey`, and derive `npmDependencies` from it. For example, in the plugin:
```scala
npmDependencies ++= otherKey.value.map(_._1)
```
and in the build:
```scala
otherKey += ("ace-builds", "1.2.9") -> List(...)
```
Note that this even highlights that `otherKey` could be part of a *third* plugin, which depends on `NpmDepsPlugin`, maybe called `NpmDepsExecutionFilesPlugin`.

In any case, it leaves the opportunity to use `NpmDepsPlugin` with `npmDependencies` only, without altering the execution files (which is what `ScalaJSBundlerPlugin` wants).

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
sjrd: Why don't you use `(packageMinifiedJSDependencies in Compile).value` in your demo instead, then?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
sjrd: Uh, well, if you *cut*-and-paste it from `ScalaJSBundlerPlugin` (which you should if `NpmDepsPlugin` factors out stuff from `ScalaJSBundlerPlugin` instead of living beside it), then keep it, as the PR doesn't make the situation worse.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
mathieuleclaire: To render the API smarter. Here, you only have to call ```dependencyFile.value```, and not both:
```scala
skip in packageJSDependencies := false
...
packageMinifiedJSDependencies in Compile).value
```

sjrd: In the current definition of the plugin, `dependencyFile` does not influence `skip in packageJSDependencies`, so anyway you need two "things":
```scala
skip in packageJSDependencies := false
...
dependencyFile.file
```
I do not see the added value.

mathieuleclaire: Yes, but it is hidden in the NpmDepsPlugin project settings. So that, you only have one line to write in your project definition and not 2.

sjrd: My comment here does not suggest removing `skip in packageJSDependencies := false` from the plugin. If you just remove the `dependencyFile` from the sbt plugin, then the only thing that changes in the demo is to replace `dependencyFile.value` by `(packageMinifiedJSDependencies in Compile).value`. It is still one line in the project definition.

mathieuleclaire: You're right. On the other hand, isn't it a little bit confusing since these two lines are still use together in scalajs ?

mathieuleclaire:
But, npmDependencies is actually not used in NpmDepsPlugin

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
sjrd: > On the other hand, isn't it a little bit confusing since these two lines are still use together in scalajs ?

The whole settings around `jsDependencies` is confusing because it's mostly broken, and that's why it's been extracted away so that people who don't need it can ignore it.

sjrd: No. I recommend that `NpmDepsPlugins` only takes care of `npmDependencies`, `npmUpdate`, and immediate friends that do not depend on `jsDependencies` (e.g., `useYarn`, `packageJson`).

Then, you could start a new project that links `NpmDepsPlugins` with `jsDependencies`, `resolvedJSDependencies`, and friends (e.g., `packageMinifiedJSDependencies).

`NpmDepsPlugin` would be the common denominator of `ScalaJSBundlerPlugin` and your other plugin.

Less magic, less responsibilities per plugin -> win/win for everyone, because everyone can choose how much they need.

sjrd: I think it should be a very different project, so that the scalajs-bundler repo can eventually be rid of the dependency on (what's been moved to) sbt-jsdependencies.

That said, you can probably *develop* it first in the same repo so that you can iterate more quickly, but then PR only the part for `NpmDepsPlugin`.

mathieuleclaire:
So, I kept NpmDepsPlugin as a generic plugin implementing npmUpdate and I moved my particular keys in a new separate plugin depending on NpmDepsPlugin: https://github.com/ISCPIF/scalajs-execnpm. What do you think ?

build.sbt
```diff
@@ -8,6 +8,7 @@ val `sbt-scalajs-bundler` =
     .settings(commonSettings)
     .settings(
       sbtPlugin := true,
+      version := "0.11.0",
```
sjrd: Not sure how @julienrf manages versions, but this should probably not be here.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
```diff
@@ -0,0 +1,107 @@
...
+    TaskKey[BundlerFile.PackageJson]("scalaJSBundlerPackageJson",
+      "Write a package.json file defining the NPM dependencies of project",
+      KeyRanks.Invisible
+    )
+
+  override lazy val projectSettings =
+    perScalaJSStageSettings(fullOptJS) ++ perScalaJSStageSettings(fastOptJS)
+
+  protected def perScalaJSStageSettings(stage: TaskKey[Attributed[File]]): Seq[Def.Setting[_]] = Seq(
+    useYarn := false
```
sjrd: In `ScalaJSBundlerPlugin`, `useYarn` used to be set in the project scope, directly in `projectSettings`. I think that's also where it should be in `NpmDepsPlugin`.

Speaking of which, the `useYarn := false` should be removed from `ScalaJSBundlerPlugin.projectSettings`, since it's redundant with the assignment in `NpmDepsPlugin.projectSettings`.

sjrd: In fact, you should inline the contents of your current `persScalaJSStageSettings` inside `projectSettings`, since you're not using `stage` at all here (which makes sense). So currently, you're just duplicating all these settings by calling `perScalaJSStageSettings(fullOptJS) ++ perScalaJSStageSettings(fastOptJS)`, for no reason.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
```diff
@@ -0,0 +1,107 @@
...
+  ) ++ inConfig(Compile)(perConfigSettings)
+
+
+  private lazy val perConfigSettings: Seq[Def.Setting[_]] = Seq(
+    npmUpdate := {
+      val log = streams.value.log
+      val targetDir = (crossTarget in npmUpdate).value
+      val jsResources = scalaJSNativeLibraries.value.data
+      val packageJsonFile = scalaJSBundlerPackageJson.value
```
sjrd: You're using `scalaJSBundlerPackageJson.value` from `NpmDepsPlugin`, but `NpmDepsPlugin` does not assign any value to that task. I'm not sure how `NpmDepsPlugin` can be used on its own under those circumstances.

You should probably add a scripted test with a project that only uses `NpmDepsPlugin` (but not `ScalaJSBundlerPlugin`), and does something meaningful with it. You might uncover other similar issues.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
```diff
@@ -0,0 +1,107 @@
...
+      "Write a package.json file defining the NPM dependencies of project",
+      KeyRanks.Invisible
+    )
+
+  override lazy val projectSettings =
+    perScalaJSStageSettings(fullOptJS) ++ perScalaJSStageSettings(fastOptJS)
+
+  protected def perScalaJSStageSettings(stage: TaskKey[Attributed[File]]): Seq[Def.Setting[_]] = Seq(
+    useYarn := false
+  ) ++ inConfig(Compile)(perConfigSettings)
```
sjrd: You also need `inConfig(Test)(perConfigSettings)` here, otherwise `npmUpdate` cannot be used in the `Test` configuration.

build.sbt
mathieuleclaire: Yes, I have to correct this, it was for my tests.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
mathieuleclaire: scalaJSBundlerPackageJson is set in ScalaJSBundlerPlugin ```perConfigSettings``` and set it in my ExecNpmPlugin ```perConfigSettings```

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
mathieuleclaire: I commited my modifications except for the scalaJSBundlerPackageJson. Sadly I was off the previous 15 days ;) Do you think you and @julienrf that a 0.10.1 could be published with this PR, once it will be OK ?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
sjrd: That means that `NpmDepsPlugin` is not useful on its own, so it is probably not a good delineation of an abstraction. `NpmDepsPlugin` should define `scalaJSBundlerPackageJson` itself (and it should be called `npmDepsPackageJson`), in a way that is generic enough to be reused.

From the way it is defined in `ScalaJSBundlerPlugin`, I don't see why the definition there cannot be factored out (almost) as is into `NpmDepsPlugin`. The only bundler-specific things there are `(version in webpack).value` and `(version in startWebpackDevServer).value`, which means that those two things need an additional hook in `NpmDepsPlugin`. Other keys such as `npmDependencies` can be extracted into `NpmDepsPlugin` as well. It makes sense. Then in your plugin `ExecNpmPlugin` you could automatically derive `npmDependencies` from your `allNpmDeps`, and leave the other ones alone since they would have sensible defaults. Doing it this way will result in even less code in `ExecNpmPlugin`, which kind of proves that it would provide a better abstraction.

mathieuleclaire: I didn't define ```scalaJSBundlerPackageJson``` in NpmDepsPlugin because I don't want it to depend on ```webpack```. I redefine ```scalaJSBundlerPackageJson``` with fake values in my ```ExecNpmPlugin```. So, what should I do ? Define two new settingsKey for both versions at the ```NpmDepsPlugin``` level (so that their values are set in ```scalaJSBundlerPlugin``` with the webpack version values and with fake ones in the ```ExecNpmPlugin```) ?

sjrd: > Define two new settingsKey for both versions at the NpmDepsPlugin level (so that their values are set in scalaJSBundlerPlugin with the webpack version values and with fake ones in the ExecNpmPlugin) ?

Yes, I think that's what is needed, here.

sjrd: Or perhaps better, remove `webpackVersion` and `webpackDevServerVersion` from the APIs of `PackageJson` altogether. In that case, `ScalaJSBundlerPlugin` would automatically add the relevant pairs to the `npmDevDependencies`, as is currently done [here](https://github.com/scalacenter/scalajs-bundler/blob/e089cf44e0066e484a8799cc35019ea8a406ff14/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala#L42-L59).

So `ScalaJSBundlerPlugin.perConfigSettings` would contain something like
```scala
npmDevDependencies ++= {
  val webpackVersion := (version in webpack).value
  val webpackDevServerVersion := (version in startWebpackDevServer).value

   val sourceMapLoaderVersion = 
      NpmPackage(webpackVersion).major match {
        case Some(1) | Some(2) => "0.1.5"
        case Some(3) => "0.2.1"
        case Some(x) => sys.error(s"Unsupported webpack major version $x")
        case None => sys.error("No webpack version defined")
      }

    npmDevDependencies.value ++ Seq(
      "webpack" -> webpackVersion,
      "webpack-dev-server" -> webpackDevServerVersion,
      "concat-with-sourcemaps" -> "1.0.4", // Used by the reload workflow
      "source-map-loader" -> sourceMapLoaderVersion // Used by webpack when emitSourceMaps is enabled
    )
}
```
(more or less)

and then `NpmDepsPlugin` wouldn't have to even know about `webpackVersion` or `webpackDevServerVersion`. And you wouldn't need any fake version for those in your plugin. :)

mathieuleclaire: Yes, but I still have to define ```npmDependencies``` and stuff in NpmDepsPlugin, which does not really make sense. 
Actually I don't need to have ```devDependencies``` to be written at all in my package.json. Why not having a ```writePackageJson``` function with optional ```devDependencies``` arguments(```npmDependencies.value```, ```npmDevDependencies.value```, ```npmResolutions.value```,```additionalNpmConfig.value```, and both webpack versions) ?

sjrd: Why would `npmDependencies` not make sense in `NpmDepsPlugin`!? On the contrary, it makes *perfect* sense. It's even the name of the plugin: *NpmDeps*Plugin.

And the other arguments are already "optional" in the sense that they are collections, and their default values are to be empty. No need to make them more optional than that.

mathieuleclaire: > Why would npmDependencies not make sense in NpmDepsPlugin!?

Deps is not the same than ```Dependencies```, it is a collection of ``` case class Dep(module: String, version: String, jsFiles: List[String])```. But now it has been moved in ```ExecNpmPlugin```. So ```NpmDepsPlugin``` should be renamed to avoid confusion. Something like NpmUpdatePlugin.

And, you right, adding Options is ugly. 
Well, it all about 3 lines of json to be added or not ! :)

mathieuleclaire: I splited the write function of package.json into devDependencies and depencies parts. The ```PackageJson.write``` method is still called from ```ScalaJSBundlerPlugin``` and ```PackageJson.writeDepsOnly``` is called from my ```ExecNpmPlugin```. This way, no more fake values and partial fake package.json file are generated.

sjrd:
I don't understand where we're going anymore. The way it's done currently, `NpmUpdatePlugin` is *just a function*, and would be advantageously replaced by precisely that: just a standalone `def`, no plugin whatsoever.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmDepsPlugin.scala
sjrd: > Deps is not the same than Dependencies, it is a collection of case class Dep(module: String, version: String, jsFiles: List[String]).

My point all along has been that you can *derive* `npmDependencies` from what your `Dep`s. And that therefore, `NpmDepsPlugin` (which should perhaps be called `NpmDependenciesPlugin`, to avoid naming clashes) can be implemented in terms of `npmDependencies`, and that your plugin can automatically derive `npmDependencies` in terms of what you call `npmDeps`.

If we're not going at least *that* far in factoring out things out of `ScalaJSBundlerPlugin`, then `NpmUpdatePlugin` is really *just* a `def`, and should not be an sbt plugin at all.

mathieuleclaire:
What ```def``` ? The ```npmUpdate``` ? I use ```npmUpdate``` and ```writeDepsOnly``` ? I don't understand how could use it as an function ?

sjrd:
`npmUpdate`. It really is just a tiny function, with some inputs and some side effects and outputs. If the inputs do not receive sensible defaults from other parts of the sbt plugin as other settings and tasks, then the sbt plugin is not providing any feature beside having a `def npmUpdate(inputs): Outputs`.

Note that it is entirely possible that it is the right thing to do for your use case. In that case, it may be better to revert to the original state, and just expose this `def npmUpdate` as a part of the non-sbt API, which you can call from your plugin.

mathieuleclaire:
You are right ! I removed the ```NpmUpdatePlugin``` and I defined npmUpdate in a ```NpmUpdateTasks``` object. All in all, I have two new function definitions: ```npmUpdate``` and ```writeOnlyDepsPackageJson```. My ```ExecNpmPlugin``` works fine with these 2 new functions. Is it now acceptable for you ?

mathieuleclaire:
@julienrf : is it fine for you ?

julienrf:
@mathieuleclaire A lot of discussion has happened around this PR, is the PR description up to date? Can you guide us to review the PR?

mathieuleclaire:
Not exactly. @sjrd finally said my plugin should be external to scalajs-bundler. But I need to call 2 adapded functions: 
- npmUpdate in a separate object: https://github.com/mathieuleclaire/scalajs-bundler/blob/0862712bc8714ac29d1a9ba260450a79dab080b0/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala#L15:L49 and 
- writeOnlyDepsPackageJson: https://github.com/mathieuleclaire/scalajs-bundler/blob/0862712bc8714ac29d1a9ba260450a79dab080b0/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/PackageJsonTasks.scala#L79:L111

It has no effect on the scalajs-bundler behaviour (just calling npmUpdate in a separate object) and it is very helpfull for me since I can use npmUpdate out of the scalajs-bundler plugin context.

julienrf:
What do you think of the initial idea of having a plugin on which `ScalaJSBundlerPlugin` depends? That plugin would define the `npmDependencies` and `npmDevDependencies` keys and expose a task to write a `package.json` file. Then, `ScalaJSBundlerPlugin` would add it’s own entries to the `npmDevDependencies` key. And we would reuse the same method to generate the `package.json` file.

mathieuleclaire:
Well, we are talking since months about this issue. I changed my implementations few times according to what you wanted (create a third plugin inside scalajs-bundler, then implement separate functions, then implement an external plugin). My only goal was not to duplicate scalajs-bundler functions. Anyway. 

Could you at least accept [the npmUpdate function in a separate object](https://github.com/mathieuleclaire/scalajs-bundler/blob/0862712bc8714ac29d1a9ba260450a79dab080b0/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala#L15:L49), so that it can be used in external projects ?

Thanks

julienrf:
Yes.

mathieuleclaire:
How do we proceed ?

julienrf:
It’s up to you, you can update that PR or start a new one, it doesn’t really matter for me.

mathieuleclaire:
Here we are !

julienrf:
Sorry to bother you again but can you remove the changes you’ve applied to code style only, and squash your commits into a single one?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
```diff
@@ -0,0 +1,47 @@
...
+import scalajsbundler.BundlerFile.PackageJson
+
+object NpmUpdateTasks {
+
+  def npmUpdate(targetDir: File,
+                packageJsonFile: File,
+                useYarn: Boolean,
+                jsResources: Seq[VirtualJSFile with RelativeVirtualFile],
+                streams: Keys.TaskStreams
+                ) = {
```
julienrf: Would you mind adding some scaladoc?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -47,7 +47,7 @@ import scalajsbundler.util.JSON
   * Version of webpack-dev-server to use.
   *
   * {{{
-  *   version in startWebpackDevServer := "2.7.1"
+  *   version in startWebpackDevServer := "2.11.1"
```
julienrf: Why this change?

mathieuleclaire: No idea :). My mistake.

mathieuleclaire:
How could I squash all my commits into a single one ? By reverting and recommiting ?

julienrf:
Nevermind, I can squash them by myself if you are not familiar with that process. In that case, just make sure that the PR only contains the minimum amount of changes (ie no code style changes).

mathieuleclaire:
I hope it is fine now :) !

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -324,6 +324,18 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
       taskKey[Seq[File]]("Files that trigger webpack launch")
 
+    /**
+      * Additional arguments to webpack
+      *
+      * Defaults to an empty list.
+      *
+      * @group settings
+      */
+    val webpackExtraArgs = SettingKey[Seq[String]](
```
julienrf: This key already exists in master. So, you should rebase your work on top of master.

In the end, it’s probably simpler to start a fresh PR on top of master.

PR #210, opened 2018-01-29
Bundler doesn't support version of jsdom more than 9.

PepRoll:
I installed JSDOM 11.6.1. When I started the test. I got this error.

```
jsdom.createVirtualConsole();
      ^
TypeError: jsdom.createVirtualConsole is not a function
```
This problem relates to this [issue](https://github.com/scala-js/scala-js/issues/2902)


julienrf:
Thanks!

PR #213, opened 2018-02-03
Fix #168: Update the snabbdom facade for Scala.js 1.0

### LINKED ISSUE #168: Update the snabbdom facade for Scala.js 1.0 in the tests
From the CI:

~~~
[info] [warn] /tmp/sbt_3bdc13c/play/client/src/main/scala/snabbdom/snabbdom.scala:23: Top-level native JS classes and objects should have an @JSGlobal or @JSImport annotation. This will be enforced in 1.0.
[info] [warn]   If migrating from 0.6.14 or earlier, the equivalent behavior is an @JSGlobal without parameter.
[info] [warn]   (you can suppress this warning in 0.6.x by passing the option `-P:scalajs:suppressMissingJSGlobalDeprecations` to scalac)
[info] [warn] class VNode(
[info] [warn]       ^
[info] [warn] /tmp/sbt_3bdc13c/play/client/src/main/scala/snabbdom/snabbdom.scala:33: Top-level native JS classes and objects should have an @JSGlobal or @JSImport annotation. This will be enforced in 1.0.
[info] [warn]   If migrating from 0.6.14 or earlier, the equivalent behavior is an @JSGlobal without parameter.
[info] [warn]   (you can suppress this warning in 0.6.x by passing the option `-P:scalajs:suppressMissingJSGlobalDeprecations` to scalac)
[info] [warn] class VNodeData extends js.Object
[info] [warn]       ^
~~~

We should make these classes non top-level or add them a `@JSImport` annotation.

exoego:
This PR changes `VNode` and `VNodeData` from class to trait, since [their JS/TS counterpart are interface](https://github.com/snabbdom/snabbdom/blob/master/src/vnode.ts#L13) therefore those are NOT class. 

The warning for those facades (`JS classes and objects should have an @JSGlobal or @JSImport annotation`) will be resolved, since those 2 types become trait.

If we add `@JSImport` to them, as mentioned in #168, making an instance (`new VNode(...)`) will fail since they does not have constructor in original JS/TS. So I thinks it is better to change them to trait.

### DIFF
--- a/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/static-webpack2/src/main/scala/snabbdom/snabbdom.scala
+++ b/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/static-webpack2/src/main/scala/snabbdom/snabbdom.scala
@@ -20,16 +20,14 @@ object h extends js.Function3[String, js.UndefOr[js.Any], js.UndefOr[js.Any], VN
 }
 
 @js.native
-class VNode(
-  selector: js.UndefOr[String],
-  data: js.UndefOr[VNodeData],
-  children: js.UndefOr[js.Array[VNode | String]],
-  text: js.UndefOr[String],
-  elm: js.UndefOr[Element | Text],
-  key: js.UndefOr[String | Double]
-) extends js.Object
+trait VNode extends js.Object {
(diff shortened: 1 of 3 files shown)


julienrf:
Thanks @exoego !

PR #214, opened 2018-02-03
Fix #181: Migrate from requiresDOM to JSDOMNodeJSEnv

### LINKED ISSUE #181: deprecate requiresDOM in Test := true
Like mentioned in gitter, sbt prints:
```
 warning: value requiresDOM in object AutoImport is deprecated: Requesting a DOM-enabled JS env with `jsDependencies += RuntimeDOM` or `requiresDOM := true` will not be supported in Scala.js 1.x. Instead, explicitly select a suitable JS with `jsEnv`, e.g., `jsEnv := new org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv`.
    requiresDOM in Test := true,
```

But when removing this line, the tests fail with node saying "cannot find module 'jsdom'".

exoego:
This PR fixes https://github.com/scalacenter/scalajs-bundler/issues/181

### DIFF
--- a/manual/src/ornate/reference.md
+++ b/manual/src/ornate/reference.md
@@ -44,7 +44,7 @@ your Scala facades (you can see an example
 If your tests execution environment require the DOM, add the following line to your build:
 
 ~~~ scala
-requiresDOM in Test := true
+jsEnv := new org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
 ~~~
 
 Then, `ScalaJSBundlerPlugin` will automatically download jsdom and bundle the tests before
--- a/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace-with-jsdom-unit-testing/build.sbt
+++ b/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace-with-jsdom-unit-testing/build.sbt
@@ -25,7 +25,7 @@ webpackConfigFile in fastOptJS := Some(baseDirectory.value / "dev.webpack.config
 webpackConfigFile in Test := Some(baseDirectory.value / "test.webpack.config.js")
 
 // Execute the tests in browser-like environment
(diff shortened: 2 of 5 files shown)


julienrf:
I’m surprised that it works because we do rely on `scalaJSRequestsDOM` in our code: https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala#L611

Maybe our test suite is not strong enough. I’ll have a deeper look on Monday.

julienrf:
@exoego We still use `scalaJSRequestsDOM` in `ScalaJSBundlerPlugin`. We should update its usage according to these instructions:
https://github.com/scala-js/scala-js/blob/0.6.x/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L120-L129
(however I don’t know how we are supposed to check whether a `JSEnv` is “DOM-enabled” or not, do you have any pointer, @sjrd?)

Also, if jsdom tests turn out to work without all this machinery, then we can just remove it instead of migrating it…

sjrd:
> however I don’t know how we are supposed to check whether a JSEnv is “DOM-enabled” or not

You cannot. I suggest you define and use a custom setting in scalajs-bundler if you need to know that.

cquiroz:
Any chance of reviving this? I'd love to remove my 
`requiresDOM in Test`

exoego:
I am not sure how to proceed this PR.
Any guidance is appreciated.

ngbinh:
see https://github.com/scalacenter/scalajs-bundler/pull/266

exoego:
So should I close this PR in favor of PR https://github.com/scalacenter/scalajs-bundler/pull/266 ?

julienrf:
Hey, indeed I think #266 will do the job but thanks a lot for having initiated the work!

PR #215, opened 2018-02-05
Fix #99: Relax NPM dependency version conflict resolution

### LINKED ISSUE #99: Relax NPM dependency version conflict resolution
Current implementation requires that for each NPM package all version descriptors match exactly (for example, `15.4.2` and `~15.4.2` don't match). So any such "conflict" must be resolved explicitly via `npmResolutions`. This is not a big problem at the moment, but is likely to become more of an issue as the number of libraries using the bundler grows.
There are two ways to deal with this I can think of:
1. Automatic resolution for such conflicts within the plugin itself. This would, however, require adding and maintaining conflict resolution code compatible with `package.json` specification.
2. A much simpler solution would be to delegate such resolution to `npm`/`yarn` themselves. `package.json` supports passing multiple version descriptors, which should be perfect for this purpose.

exoego:
This PR fixes https://github.com/scalacenter/scalajs-bundler/issues/99 by delegating version conflict resolution to `npm` or `yarn`, if `npmDependencies` is not provided for a conflicted package.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageJson.scala
@@ -99,8 +99,10 @@ object PackageJson {
               case Seq(single) =>
                 name -> single
               case _ =>
-                val resolution =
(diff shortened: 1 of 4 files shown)


julienrf:
Thanks @exoego!

PR #218, opened 2018-02-11
Fix #136: Add more precise jsdom detection.

### LINKED ISSUE #136: Detection of jsdom installation is not precise enough
We just check if there is a `target/scalajs-bundler-jsdom` directory, but we should check for the presence of `target/scalajs-bundler-jsdom/node_modules/jsdom`.

exoego:
This PR fixes https://github.com/scalacenter/scalajs-bundler/issues/136


julienrf:
Thanks!

PR #220, opened 2018-02-20
Support passing an extra list of arguments to webpack

cquiroz:
I think it would be useful to be able to pass custom args to webpack so I added this. I'm not very familiar with sbt plugins nor webpack but with this change I'm able to e.g. pass "--profile" which has helped me to find bottlenecks

If you think this is useful I'd add documentation

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
@@ -136,6 +136,7 @@ object Webpack {
      webpackResources: Seq[File],
      entry: BundlerFile.Application,
      targetDir: File,
+     extraArgs: Seq[String],
      log: Logger
   ): BundlerFile.ApplicationBundle = {
     writeConfigFile(emitSourceMaps, entry, generatedWebpackConfigFile, None, log)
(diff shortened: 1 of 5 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
```diff
@@ -145,7 +146,8 @@ object Webpack {
       .getOrElse(generatedWebpackConfigFile.file)
 
     log.info("Bundling the application with its NPM dependencies")
-    Webpack.run("--config", configFile.absolutePath)(targetDir, log)
+    val args = extraArgs ++: Seq("-config", configFile.absolutePath)
```
julienrf: I guess there is a typo here, there should be two dashes before “config”, right?

cquiroz: Fixed now, I forgot to upload the fix

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
```diff
@@ -8,6 +8,7 @@ import scala.sys.process.ProcessLogger
 object Commands {
 
   def run(cmd: Seq[String], cwd: File, logger: Logger): Unit = {
+    logger.debug(s"Command: ${cmd.mkString(" ")}")
```
cquiroz: This is not strictly part of the change but I found this very useful for debugging

julienrf:
Thanks!

PR #221, opened 2018-02-21
Documentation for webpackExtraArgs

cquiroz:
Just a little extra after #220 to document the new property

### DIFF
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -280,3 +280,21 @@ reload on every change:
 ~~~
 webpackDevServerExtraArgs := Seq("--inline")
 ~~~
+
+## How to pass extra parameters to webpack
+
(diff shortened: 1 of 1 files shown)


manual/src/ornate/cookbook.md
```diff
@@ -280,3 +280,14 @@ reload on every change:
...
+## How to pass extra parameters to webpack
+
+You can add extra params to be passed to webpack, e.g. to increase debugging
+
+~~~
+webpackExtraArgs := Seq("--profile", "--progress", "true")
+~~~
+
+**Note** Params are passed verbatim, they are not sanitized and could create errors when passed to webpack.
+In particular avoid setting the "--config" param.
```
julienrf: I would rather explain at the beginning what the plugin does when it invokes webpack (= it sets the `--config` option to pick the right configuration file), but it is possible to set extra arguments with `webpackExtraArgs` (to set something *else* than `--config`).

cquiroz:
Redacted as suggested and rebased onto master

PR #222, opened 2018-03-02
Make library example work in OSX

cquiroz:
The library example doesn't work in OSX, as there is a clash on the names `uuid` and `UUID`. This is due to osx's file system isn't case sensitive by default

### DIFF
--- a/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/library/src/main/scala/example/Library.scala
+++ b/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/library/src/main/scala/example/Library.scala
@@ -11,8 +11,8 @@ object Library {
 //#library-definition
 
 object SomeOtherCode {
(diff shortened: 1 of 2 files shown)


julienrf:
Not sure what the clash is about because there is only one `uuid.scala` file, right?

cquiroz:
It happens because there is a trait `UUID` and an object `uuid` which both produce class files with the same name (uuid.sjsir) and in osx they overwrite each other

cquiroz:
The original reason for this PR is this error when running `master` on OSX

```
[info] [error] (compile:compileIncremental) java.lang.IllegalArgumentException: requirement failed: Source file '/Users/cquiroz/Projects/scalajs-bundler/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/library/target/scala-2.11/classes/uuid/uuid.sjsir' does not exist.
```

julienrf:
Thanks for the explanation!

PR #225, opened 2018-03-08
Add --mutex to yarn command

fdietze:
Yarn cannot be run in parallel and therefore needs a mutex:
https://yarnpkg.com/lang/en/docs/cli/#toc-concurrency-and-mutex

I ran into this when when building two subprojects in sbt which both use scalajs-bundler.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
@@ -44,7 +44,7 @@ object ExternalCommand {
     */
   def install(installDir: File, useYarn: Boolean, logger: Logger)(npmPackages: String*): Unit =
     if (useYarn) {
-      Yarn.run("add" +: "--non-interactive" +: npmPackages: _*)(installDir, logger)
+      Yarn.run("add" +: "--mutex" +: "file:/tmp/.yarn-mutex" +: "--non-interactive" +: npmPackages: _*)(installDir, logger)
     } else {
       Npm.run("install" +: npmPackages: _*)(installDir, logger)
     }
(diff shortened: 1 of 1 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
```diff
@@ -44,9 +44,9 @@ object ExternalCommand {
     */
   def install(installDir: File, useYarn: Boolean, logger: Logger)(npmPackages: String*): Unit =
     if (useYarn) {
-      Yarn.run("add" +: "--non-interactive" +: npmPackages: _*)(installDir, logger)
+      Yarn.run("add" +: "--mutex" +: "file:/tmp/.yarn-mutex" +: "--non-interactive" +: npmPackages: _*)(installDir, logger)
```
julienrf: I guess this is not cross platform?

We should probably get passed the `yarnMutexFile` and generate it from sbt.

fdietze: Good point!

sjrd: Or just use `Yarn.synchronized { ... }` around that piece of code? Only works for multiple calls within the same sbt process, though.

fdietze: Maybe it's simpler to use the network-mutex: `--mutex network`. Will this work on windows?

julienrf:
Thanks @fdietze!

dispalt:
Any reason we couldn't do this to the yarn install command?

dispalt:
And to be clear I am talking about here https://github.com/scalacenter/scalajs-bundler/blob/1310f87f64da19afaf0e5d3025a5c106d09475bb/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala#L32

fdietze:
Oh, you are right. I overlooked that one.

PR #232, opened 2018-03-22
Example with a custom test config

cquiroz:
I thought it could be useful to add an example with a custom test config. This example shows:

* use webpack 4.
* use utest to run tests.
* example of using react and react test.
* differentiate the webpack configuration for tests.


julienrf:
Thanks @cquiroz (and sorry for the delay… I’ve put it in my queue and came back to it only now…)

Issue #192, opened 2017-11-30
Webpack failed to create application bundle


mathieuleclaire:
Hi,
I am packaging some facade with scalajs-bundler. I added some config-files (to avoid using jquery): 
```js
module.exports = {
      "entry": {
    "demo-opt": [__dirname + "/bootstrapslider-opt.js"]
  },
  "output": {
    "path": __dirname,
    "filename": "[name]-bundle.js"
  },
    resolve: {
        alias: {
             "jquery": __dirname + "../../../resources/jquery-stub.js"
        }
    }
}
```
my built.sbt project is the following: 
```sbt
lazy val bootstrapslider = project.in(file("bootstrapslider")) enablePlugins (ScalaJSBundlerPlugin) settings(
  scalaJsDom,
  querki,
  npmDependencies in Compile += "bootstrap-slider" -> bootstrapSliderVersion,
  jsDependencies += ProvidedJS / "jquery-stub.js",
  webpackConfigFile := Some(baseDirectory.value / "jqueryConfig.config.js")
)
```

on a ```fullOptJS::wepack```, I get:

```sh
sbt:bootstrapslider> fullOptJS::webpack
[info] Writing scalajs.webpack.config.js
[info] Bundling the application with its NPM dependencies
[info] Hash: 8624608efa0a54f8ae83
[info] Version: webpack 3.5.5
[info] Time: 76ms
[info]              Asset     Size  Chunks             Chunk Names
[info] demo-opt-bundle.js  12.6 kB       0  [emitted]  demo-opt
[info]    [0] multi ./bootstrapslider-opt.js 28 bytes {0} [built]
[info]    [1] ./bootstrapslider-opt.js 9.27 kB {0} [built]
[info]    [2] (webpack)/buildin/global.js 509 bytes {0} [built]
[error] java.lang.AssertionError: assertion failed: Webpack failed to create application bundle
[error] 	at scala.Predef$.assert(Predef.scala:219)
[error] 	at scalajsbundler.Webpack$.bundle(Webpack.scala:152)
[error] 	at scalajsbundler.sbtplugin.WebpackTasks$.$anonfun$webpack$2(WebpackTasks.scala:45)
[error] 	at sbt.util.FileFunction$.$anonfun$cached$1(FileFunction.scala:73)
[error] 	at sbt.util.FileFunction$.$anonfun$cached$4(FileFunction.scala:147)
[error] 	at sbt.util.Difference.apply(Tracked.scala:313)
[...]

PR #233, opened 2018-04-05
Example application with a custom name

cquiroz:
This is a sample application demonstrating the issue #192 
This is happening because there is a custom name on the webpack configuration. Thus `scalajs-bundler` fails. The test will pass as we are expecting the failure

I'm working on parsing webpack stats to fix this problem

### DIFF
--- a/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-assets/build.sbt
+++ b/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-assets/build.sbt
@@ -1,0 +1,41 @@
+import com.gargoylesoftware.htmlunit.WebClient
+import com.gargoylesoftware.htmlunit.WebConsole.Logger
+
+name := "webpack-assets"
+
+enablePlugins(ScalaJSBundlerPlugin)
+
+scalaVersion := "2.12.5"
+
+scalaJSUseMainModuleInitializer := true
+
+//webpackBundlingMode := BundlingMode.LibraryAndApplication()
+
+// Use a custom config file
+webpackConfigFile := Some(baseDirectory.value / "webpack.config.js")
+
+npmDevDependencies in Compile += "html-webpack-plugin" -> "3.0.6"
+
+npmDevDependencies in Compile += "webpack-merge" -> "4.1.2"
+
+webpackDevServerPort := 7357
+
+version in webpack                     := "4.3.0"
+
+version in startWebpackDevServer       := "3.1.1"
+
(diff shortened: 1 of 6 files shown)


sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-assets/webpack.config.js
```diff
@@ -0,0 +1,12 @@
+const ScalaJS = require("./scalajs.webpack.config");
+const Merge = require("webpack-merge");
+const HtmlWebpackPlugin = require("html-webpack-plugin");
+
+const WebApp = Merge(ScalaJS, {
+  output: {
+    filename: "app.js"
```
cquiroz: The custom name triggers the error

cquiroz:
yes, I'm working on a bugfix, parsing the stats output with a json parser and trying to match the input file to the actual output.

It has a few side effects though. One of them is that we lose the regular output which is quite useful.

I actually want to later extend that to get a list of all the assets produced by webpack

julienrf:
Yes, getting a list of the assets produced by webpack is the right thing to do!

cquiroz:
I expect to expand this example to have other assets, e.g. images and export them to `sbt`

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-assets/build.sbt
```diff
@@ -0,0 +1,41 @@
...
+// Check that a HTML can be loaded (and that its JavaScript can be executed) without errors
+InputKey[Unit]("html") := {
+  import complete.DefaultParsers._
+  import scala.sys.process._
+
+  val page = (Space ~> StringBasic).parsed
+  import com.gargoylesoftware.htmlunit.WebClient
+  val client = new WebClient()
+  try {
+    val scalajsBundleDir = s"${baseDirectory.value.absolutePath}/target/scala-2.12/scalajs-bundler/main"
```
cquiroz: Is there a better way to refer to the dir where scala js-bundle writes the files?

julienrf: `npmUpdate in Compile`, IIRC.

cquiroz: Yeah, that works. I updated and rebase the code

cquiroz:
I'm closing this. The example is now included in #234

PR #237, opened 2018-04-25
add function as a module example

konradst:
Added example for importing function as a module
module.exports = function () {}
like camelcase in npm


julienrf:
Thanks @konradst!

PR #238, opened 2018-04-26
Fix thread leak

eatkins:
I noticed that my tests were leaking ComJSEnvRPC threads. I tracked it
down to the code in the loadedTestFrameworks settings definition in the
ScalaJSBundlerPlugin which initializes, but does not register the
adapter in the createdTestAdapters list.

I also noticed that newTestAdapter would leak threads in the recursive
case (which probably doesn't happen often in the real world).


julienrf:
Thanks @eatkins !

PR #239, opened 2018-04-28
Bugfix parsing errors on the output

cquiroz:
While testing I've found some cases not properly handled with the changes on # 234. In this case, the webpack refers to an unknown module and the error goes to stdout interfering with the output parsing.

An example is included where the webpack config refers to `less-loader` which hasn't been included as a dependency. Before this patch the error message wasn't displayed correctly and sbt wouldn't complete

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
@@ -1,12 +1,14 @@
 package scalajsbundler
 
 import sbt._
-
 import scalajsbundler.util.{Commands, JS}
 import org.scalajs.core.tools.linker.StandardLinker.Config
 import java.io.InputStream
+
 import play.api.libs.json._
 import Stats._
+
+import scala.util.{Failure, Success, Try}
 
 object Webpack {
   // Represents webpack 4 modes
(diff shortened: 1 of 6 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
```diff
@@ -246,18 +248,31 @@ object Webpack {
...
-    parsed.validate[WebpackStats] match {
-      case JsError(e) =>
-        logger.error("Error parsing webpack stats output")
-        // In case of error print the result and return None. it will be ignored upstream
-        e.foreach {
-          case (p, v) => logger.error(s"$p: ${v.mkString(",")}")
-        }
+  private def jsonOutput(cmd: Seq[String], logger: Logger)(in: InputStream): Option[WebpackStats] = {
+    Try {
+      val parsed = Json.parse(in)
```
cquiroz: I probably didn't understand `play-json` well enough. `Json.parse` can throw and that leaves the system in an unstable state

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
```diff
@@ -8,13 +8,17 @@ import scala.sys.process.ProcessLogger
 
 object Commands {
 
-  def run[A](cmd: Seq[String], cwd: File, logger: Logger, outputProcess: InputStream => A): Option[A] = {
-    val toErrorLog = (is: InputStream) => scala.io.Source.fromInputStream(is).getLines.foreach(msg => logger.error(msg))
+  def run[A](cmd: Seq[String], cwd: File, logger: Logger, outputProcess: InputStream => A): Either[String, Option[A]] = {
+    val toErrorLog = (is: InputStream) => {
+      scala.io.Source.fromInputStream(is).getLines.foreach(msg => logger.error(msg))
+      is.close()
```
cquiroz: Streams must be closed after being used by `Process`

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
```diff
@@ -23,15 +27,15 @@ object Commands {
     val processIO = BasicIO.standard(false).withOutput(outputCapture).withError(toErrorLog)
     val code: Int = process.run(processIO).exitValue()
     if (code != 0) {
-      sys.error(s"Non-zero exit code: $code")
+      Left(s"Non-zero exit code: $code")
```
cquiroz: This change let's the output parser execute even when `webpack` fails. The call to `sys.error` was too early in this case

PR #240, opened 2018-05-13
Update concat-with-sourcemaps

cquiroz:
Small update of `conact-with-sourcemaps`. Otherwise we get the message:

```
warning concat-with-sourcemaps@1.0.4: important fix in v1.0.6
```


julienrf:
Thanks!

PR #247, opened 2018-06-02
Assets to sbt

cquiroz:
This is a feature I've been looking for to let more easily package applications processed with scala.js and webpack. It will export all the assets produced by webpack on

`(wepback in (Compile, full/fastOptJS))value`

With this change, I can e.g. package an application using sbt-native-packager and make it include all the files referred on the application, e.g. css, images, etc

This PR also lets me remove a few calls which were a bit strange like `.asApplicationBundle(None)` when we didn't have access to the `Stats`.

The test now includes a css that is referred from the entry point and that css and its map are now included in the list of assets.

I had to do one workaround to get this to work. `sbt` caches files using `FileFunction` as a `Set[File]` so we'd need to reconstruct the `Library` or `ApplicationBundle` objects from that list. I adopted the convention that the first item is the main `js` file and the rest are assets to make this work across the cache. It may be a bit weak but it should work in general and preserves the cache. I'm using `ListSet` to preserve the order.

### DIFF
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -303,12 +303,25 @@ In particular, don't attempt to override the `--config` param.
 
 `scalajs-bundler` (version 0.12.0 onwards) supports webpack 4. To enable webpack 4, set the correct versions in `build.sbt`
 
--~~~ scala
+~~~ scala
 version in webpack := "4.8.1"
 
 version in startWebpackDevServer := "3.1.4"
--~~~
+~~~
 
 Additionally, you need to update any webpack plugins your config uses, to Webpack 4 compatible versions.
 
 Webpack 4 has the potential to substantially reduce your webpack compilation times (80% reductions have been observed but your mileage may vary)
+
+## How to use get a list of assets
+
+`scalajs-bundler` (version 0.13.0 onwards) will export a list of all assets preduced by webpack. You can read that list on sbt
+
+~~~ scala
+val files = (webpack in (Compile, fullOptJS)).value
+~~~
+
+You can use it e.g. with `[sbt-native-packager](https://github.com/sbt/sbt-native-packager)` to add mappings as:
+~~~ scala
+mappings in (Compile, packageBin) ++= (webpack in (Compile, fullOptJS)).value.map { f => f.data -> f.data.getName() },
+~~~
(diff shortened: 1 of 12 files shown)


cquiroz:
I tested this on sbt 1.x and somewhat surprisingly some tests fail on 0.13.x. I'll try to isolate them and get a fix that would work in both cases

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/LibraryTasks.scala
```diff
@@ -87,24 +87,21 @@ object LibraryTasks {
...
+            entryPointFile,
+            mode.exportedName,
+            extraArgs,
+            webpackMode,
+            log
+          ).cached
         }
-      cachedActionFunction(monitoredFiles.to[Set])
-      generatedWebpackConfigFile.asLibrary(None)
+      val cached = cachedActionFunction(monitoredFiles.to[Set])
```
julienrf: Hmm, ok anyway `cachedActionFunction` can not return a `ListSet`… It can only return a `Set`.

cquiroz: Right, we are limited by `FileFunction`'s API

cquiroz:
@julienrf All test are passing now with my last commit. I wanted to reconfirm  your approval

manual/src/ornate/cookbook.md
```diff
@@ -303,12 +303,25 @@ In particular, don't attempt to override the `--config` param.
...
+`scalajs-bundler` (version 0.13.0 onwards) will export a list of all assets preduced by webpack. You can read that list on sbt
+
+~~~ scala
+val files = (webpack in (Compile, fullOptJS)).value
+~~~
+
+You can use it e.g. with `[sbt-native-packager](https://github.com/sbt/sbt-native-packager)` to add mappings as:
+~~~ scala
+mappings in (Compile, packageBin) ++= (webpack in (Compile, fullOptJS)).value.map { f => f.data -> f.data.getName() },
+~~
```
julienrf: One “~” is missing here

julienrf:
Just fix the markdown and squash the commits and this is good to go, thanks!

PR #248, opened 2018-06-05
Updated changelog

cquiroz:
Updated for the 0.13 release

### DIFF
--- a/manual/src/ornate/changelog.md
+++ b/manual/src/ornate/changelog.md
@@ -1,17 +1,22 @@
 # Changelog
 
-## Version 0.13.0-RC1
+## Version 0.13.0
 
-> 2018 May 15
+> 2018 Jun 5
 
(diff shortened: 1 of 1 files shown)


manual/src/ornate/changelog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Version 0.13.0
+
+> 2018 Jun 5
+
+The following PRs are included in this release
+
+- [#247](https://github.com/scalacenter/scalajs-bundler/pull/247): Assets to sbt
+
```
julienrf: What do you think about replacing the 0.13.0-RC1 section instead of creating a new 0.13.0 section?

cquiroz: That's fine. I thought about doing that instead. I'll change it

cquiroz: Updated

PR #249, opened 2018-06-06
Sync yarn.lock between baseDir and installDir

### LINKED ISSUE #151: yarn.lock is not version controlled.
The yarn lock is in the target directory. It should be version controlled. I suggest it should go in the project `baseDirectory`. For example, `client/baseDirectory`.

> In order to get consistent installs across machines, Yarn needs more information than the dependencies you configure in your package.json. Yarn needs to store exactly which versions of each dependency were installed.

https://yarnpkg.com/lang/en/docs/yarn-lock

fdietze:
I was not able to test this locally...

(https://gitter.im/scalacenter/scalajs-bundler?at=5b17a736edd06f7d151380b5)

fixes #151

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
@@ -1,7 +1,8 @@
 package scalajsbundler
 
-import sbt._
+import java.io.File
 
+import sbt._
 import scalajsbundler.util.Commands
 
 /**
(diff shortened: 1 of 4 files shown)


julienrf:
Is it possible to add scripted tests for this feature?

fdietze:
Should be working correctly now! I tested it locally.

fdietze:
Any feedback on this one? I'm using it locally with my team since one week, and it's working well.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
```diff
@@ -33,20 +34,47 @@ object Npm extends ExternalCommand("npm")
...
     }
 
+  def install(baseDir:File, installDir: File, useYarn: Boolean, logger: Logger): Unit =
+    if (useYarn) {
+      syncYarnLockfile(baseDir, installDir, logger) {
+        Yarn.run("install" +: yarnOptions: _*)(installDir, logger)
+      }
+    } else {
+      Npm.run("install")(installDir, logger)
+    }
```
julienrf: Why do we need to distinguish between `yarn add` and `yarn install`, whereas this was not the case before?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
```diff
@@ -33,20 +34,47 @@ object Npm extends ExternalCommand("npm")
...
+    val targetLockFile = installDir / "yarn.lock"
+    if(sourceLockFile.exists()) {
+      logger.info("Using lockfile " + sourceLockFile)
+      IO.copyFile(sourceLockFile, targetLockFile)
+    }
+
+    yarnCommand
+
+    if(targetLockFile.exists()) {
+      logger.info("Wrote lockfile to " + sourceLockFile)
```
julienrf: Maybe put that line to the `debug` level?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
fdietze: Both, `yarn add` and `yarn install` were there before, I just moved them into `ExternalCommand`:
https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala#L47
https://github.com/scalacenter/scalajs-bundler/blob/9c3cd1b13817a471149d4552db5396c0b9849bb9/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala#L32

sbt-scalajs-bundler/src/main/scala/scalajsbundler/ExternalCommand.scala
fdietze: good idea, I'll change it.

julienrf:
Thanks!

Issue #252, opened 2018-06-28
Is it possible to ignore dependencies?


rabejens:
I am trying to build an Electron app and want to use NPM dependencies. For this, I need to `require electron`. When I run `fastOptJS::webpack` it (correctly) complains about `electron` not being there. This is correct because the `electron` module is intrinsically available when running a JS application from within Electron.

Is it possible to tell the bundler to ignore this dependency?

## Followup

I found [this post on StackOverflow](https://stackoverflow.com/questions/44663757/cannot-find-node-module-fs-when-bundling-module-via-scalajs-bundler) and added my custom `webpack.config.js` which has:

```javascript
module.exports = require('./scalajs.webpack.config');
module.exports.node = {fs: 'empty', electron: 'empty'};
```

This makes one error go away - the one about `fs` - but it still can't find `electron.

So, as a next step, I added `electron` as an NPM dependency.

Now, `fastOptJS::webpack` runs, but when I do `electron .` in the output directory, I get `TypeError: fs.existsSync is not a function`.

What am I missing?

rabejens:
I solved it. Here is how:

First, I created a `no-electron.webpack.js` in my `src/main/resources`:

```javascript
module.exports = require('./scalajs.webpack.config');
module.exports.target = 'electron';
```

Then I added to my `build.sbt`:
```scala
webpackConfigFile :=
Some(baseDirectory.value / "src" / "main" / "resources" / "no-electron.webpack.config.js")
```

Now I removed `electron` from the `npmDependencies.

When I now add a `main` to the generated `package.json` and run `electron .` in that directory, it works.

rabejens:
Yes, I would like to do that. How can I do it? Should I fork the documentation and create a PR?

PR #255, opened 2018-07-07
Added a cookbook entry on how to package for Electron

rabejens:
Added a cookbook entry on how to package for Electron as described in [Issue 252](https://github.com/scalacenter/scalajs-bundler/issues/252)

### DIFF
--- a/manual/src/ornate/cookbook.md
+++ b/manual/src/ornate/cookbook.md
@@ -192,16 +192,16 @@ dependency on the facade and to enable the `ScalaJSBundlerPlugin` plugin.
 
 ## How to use an existing facade assuming the JS library to be exposed to the global namespace? {#global-namespace}
 
-Webpack is able to require external modules by using [imports-loader](https://github.com/webpack-contrib/imports-loader) 
-and expose them to the global namespace by using [expose-loader](https://github.com/webpack/expose-loader). 
-Thus, you can write a custom webpack configuration file that uses this loaders to expose the required 
+Webpack is able to require external modules by using [imports-loader](https://github.com/webpack-contrib/imports-loader)
+and expose them to the global namespace by using [expose-loader](https://github.com/webpack/expose-loader).
+Thus, you can write a custom webpack configuration file that uses this loaders to expose the required
 modules to the global namespace. Typically, this file will look like this:
 
 ~~~ javascript src=../../../sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace-with-jsdom-unit-testing/common.webpack.config.js
 ~~~
 
 Also, tweak your `build.sbt` to add the corresponding NPM dependencies and to use the
-custom webpack configuration file: 
+custom webpack configuration file:
 
 ~~~ scala src=../../../sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/global-namespace-with-jsdom-unit-testing/build.sbt#relevant-settings
 ~~~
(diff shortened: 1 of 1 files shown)


cquiroz:
LGTM

manual/src/ornate/cookbook.md
```diff
@@ -325,3 +325,32 @@ You can use it e.g. with `[sbt-native-packager](https://github.com/sbt/sbt-nativ
 ~~~ scala
 mappings in (Compile, packageBin) ++= (webpack in (Compile, fullOptJS)).value.map { f => f.data -> f.data.getName() },
 ~~~
+
+## How to package for [Electron](https://electronjs.org)
+
+Packaging for Electron is actually not that hard either (but some more manual steps are needed to fully
+get this to work - e.g., writing Facades for Electron's built-in types.)
```
julienrf: I think we should focus on *how* to do things rather than discussing whether it’s hard or not. The part that you wrote between parens is the useful part, imho. Also, in the remaining subsections, you don’t give any clue for these facades. What about the following?

> Packaging for Electron requires to set the Webpack `target` configuration option to `electron`.

manual/src/ornate/cookbook.md
```diff
@@ -325,3 +325,32 @@ You can use it e.g. with `[sbt-native-packager](https://github.com/sbt/sbt-nativ
...
+
+### Add an extra Webpack configuration to your project
```
julienrf: I think we don’t need a subsection for this.

manual/src/ornate/cookbook.md
```diff
@@ -325,3 +325,32 @@ You can use it e.g. with `[sbt-native-packager](https://github.com/sbt/sbt-nativ
...
+
+A suggested location is `src/main/resources`. Create a file named `electron.webpack.config.js` there and add these lines:
+
+~~~ javascript
+module.exports = require('./scalajs.webpack.config');
+module.exports.target = 'electron';
+~~~
```
julienrf: Do you think you could add an electron compatible project in the tests? This way we could:

- test that the code you wrote here still works as scalajs-bundler and webpack evolve,
- insert into the documentation the code that is actually used in the test (like we do for the global namespace thing).

manual/src/ornate/cookbook.md
```diff
@@ -325,3 +325,32 @@ You can use it e.g. with `[sbt-native-packager](https://github.com/sbt/sbt-nativ
...
+Some(baseDirectory.value / "src" / "main" / "resources" / "electron.webpack.config.js")
+~~~
+
+**DON'T** add any `electron` dependencies to your `npmDependencies`!
+
+When you now run `fastOptJS` or `fullOptJS`, you get an appropriate webpack configuration
+and the generated JS files will be packaged correctly.
+
+You still have to add the generated JavaScript file to the `package.json` manually
+to have it started automatically. You could also create a build task that does this for you.
```
julienrf: Can you give an example of how to do that? Also, if you add a test project you could also add a build task that does it.

julienrf: Any advice about the Electron’s facades to write? Are some of them available somewhere?

julienrf:
[let us know](https://gitter.im/scalacenter/scalajs-bundler) if you need help to create the test project.

rabejens:
I am planning to write an example on how to create an Electron application with scalajs-bundler anyway. I’ll also have a look at the source code and see if I can add an “Electron mode” you can activate via build.sbt that does the extra steps.

PR #260, opened 2018-08-09
Add npmInstall task

### LINKED ISSUE #258: Unable to use npmUpdate in sourceGenerators
I was trying to use `npmUpdate` to get `npm` to update and also to get its location. I then wanted to use a module from there to generate some sources, specifically the apollo client for generating graphQL types.

The problem I ran into was https://github.com/sbt/sbt/issues/4292 which basically causes a cycle in the sbt tasks and it deadlocks.

The issue is that `npmUpdate` depends on a couple of things which use the `fullClasspath` and hence `exportedProduces` which eventually calls `sourceGenerators` and we get back to the start.

The first place is `scalaJSBundlerPackageJson` which seems to use the fullClasspath to find all the NPM_DEPENDENCIES. The other place is in `scalaJSNativeLibraries` to get all the `*.js` files.

I could create my own task that just calls `scalajsbundler.ExternalCommand$#install` but I was wondering if perhaps this needs better support in the plugin. Using dev dependencies during the build to generate more sources etc. seems like it would be a fairly common use case, at least in a non Scala.js world anyway.

steinybot:
This fixes #258.

I went with a more conservative approach which was to introduce a new task. We briefly discussed the removal of the staging of javascript resources from `npmUpdate` but this would have much more severe consequences for backwards compatibility.

I tested it out using https://github.com/scalacenter/scastie to make sure that the changes to `scalaJSBundlerPackageJson` to used `dependencyClasspath` didn't have any adverse affects. I first updated it to use `0.13.1` and generated the `package.json` and then updated to local build and there were no differences. I'm not sure if it is worth doing more thorough testing for this, it seems quite unlikely that change would have any negative impact, let me know what you think.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
@@ -5,13 +5,15 @@ import scalajsbundler.ExternalCommand
 import sbt._
 
 object NpmUpdateTasks {
+
   /**
-    * Runs the Npm or Yarn
-    * @param targetDir npm Directory
+    * Runs either `npm install` or `yarn install` and installs JavaScript resources as node packages.
+    *
+    * @param targetDir npm directory
     * @param packageJsonFile Json file containing NPM dependencies
     * @param useYarn Whether to use yarn or npm
-    * @param jsResources A sequence of javascript resources
-    * @param stream A sbt TaskStream
+    * @param jsResources A sequence of JavaScript resources
+    * @param streams A sbt TaskStream
     * @param npmExtraArgs Additional arguments to pass to npm
     * @param yarnExtraArgs Additional arguments to pass to yarn
     * @return The written npm directory
@@ -23,28 +25,69 @@ object NpmUpdateTasks {
                 jsResources: Seq[VirtualJSFile with RelativeVirtualFile],
                 streams: Keys.TaskStreams,
                 npmExtraArgs: Seq[String],
-                yarnExtraArgs: Seq[String]
-                ) = {
-    val log = streams.log
+                yarnExtraArgs: Seq[String]): File = {
+    val dir = npmInstallDependencies(baseDir, targetDir, packageJsonFile, useYarn, streams, npmExtraArgs, yarnExtraArgs)
+    npmInstallJSResources(targetDir, jsResources, streams)
+    dir
+  }
 
+  /**
+    * Runs either `npm install` or `yarn install`.
+    *
+    * @param targetDir npm directory
+    * @param packageJsonFile Json file containing NPM dependencies
+    * @param useYarn Whether to use yarn or npm
+    * @param streams A sbt TaskStream
+    * @param npmExtraArgs Additional arguments to pass to npm
+    * @param yarnExtraArgs Additional arguments to pass to yarn
+    * @return The written npm directory
+    */
+  def npmInstallDependencies(baseDir: File,
+                             targetDir: File,
+                             packageJsonFile: File,
+                             useYarn: Boolean,
+                             streams: Keys.TaskStreams,
+                             npmExtraArgs: Seq[String],
+                             yarnExtraArgs: Seq[String]): File = {
+    val log = streams.log
     val cachedActionFunction =
       FileFunction.cached(
-        streams.cacheDirectory / "scalajsbundler-npm-update",
+        streams.cacheDirectory / "scalajsbundler-npm-install",
         inStyle = FilesInfo.hash
       ) { _ =>
         log.info("Updating NPM dependencies")
         ExternalCommand.install(baseDir, targetDir, useYarn, log, npmExtraArgs, yarnExtraArgs)
-        jsResources.foreach { resource =>
-          IO.write(targetDir / resource.relativePath, resource.content)
-        }
         Set.empty
       }
-
-    cachedActionFunction(Set(packageJsonFile) ++
-      jsResources.collect { case f: FileVirtualJSFile =>
-        f.file
-      }.to[Set])
-
+    cachedActionFunction(Set(packageJsonFile))
     targetDir
+  }
+
+  /**
+    * Installs JavaScript resources as node packages.
+    *
+    * @param targetDir npm directory
+    * @param jsResources The JavaScript resources
+    * @param streams A sbt TaskStream
+    * @return The paths to the node packages
+    */
+  def npmInstallJSResources(targetDir: File,
+                            jsResources: Seq[VirtualJSFile with RelativeVirtualFile],
+                            streams: Keys.TaskStreams): Seq[File] = {
+    val jsFileResources = jsResources.collect {
+      case file: FileVirtualJSFile => file
+    }.toSet
+    val cachedActionFunction = FileFunction.cached(
+        streams.cacheDirectory / "scalajsbundler-npm-install-resources",
+        inStyle = FilesInfo.hash
+      ) { _ =>
+        jsFileResources.map { resource =>
+          val resourcePath = targetDir / resource.relativePath
+          IO.write(resourcePath, resource.content)
+          resourcePath
+        }
+      }
+    val files = jsFileResources.map(_.file)
+    cachedActionFunction(files).toSeq
   }
 }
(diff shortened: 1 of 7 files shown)


sjrd:
This doesn't introduce any dependency between `npmUpdate` and `npmInstall`, which means they'll duplicate work and potentially run in parallel, likely causing races. It doesn't seem to me that this is the correct solution.

julienrf:
Thanks @steinbot, I really appreciate the work you did on the documentation side! I second @sjrd, though: can we make the `npmUpdate` task reuse `npmInstall`? Also, I think we should maybe find better names… The distinction between what `npmUpdate` and `npmInstall` do is not clear just looking at their names.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
```diff
@@ -5,46 +5,115 @@ import scalajsbundler.ExternalCommand
...
                 streams: Keys.TaskStreams,
                 npmExtraArgs: Seq[String],
-                yarnExtraArgs: Seq[String]
-                ) = {
-    val log = streams.log
+                yarnExtraArgs: Seq[String]): File = {
+    val dir = npmInstallDependencies(baseDir, targetDir, packageJsonFile, useYarn, streams, npmExtraArgs, yarnExtraArgs)
+    npmInstallJSResources(targetDir, jsResources, streams)
+    dir
+  }
```
julienrf: I think you should just call `npmInstallAll` here.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -100,8 +106,82 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
       *
       * @group tasks
       */
+    @deprecated("Use npmInstall or npmInstallDependencies", "0.13.2")
```
julienrf: I think we should make it clearer that the behavior of `npmUpdate` is now implemented by the `npmInstall` task.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -562,37 +642,51 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
-      yarnExtraArgs.value),
+      npmUpdate := npmInstall.value,
+
+      crossTarget in npmInstall := (crossTarget in npmUpdate).value,
+
+      npmInstall := {
+        val dir = npmInstallDependencies.value
+        val _ = npmInstallJSResources.value
+        dir
+      },
```
julienrf: Do we need `npmInstallDependencies` to be executed *before* `npmInstallJSResources`? There is no guarantee that this will be the case. On the other hand, if we don’t need this guarantee, then we can implement `npmInstall` as follows:

~~~ scala
npmInstall := {
  val _ = npmInstallJSResources.value
  npmInstallDependencies.value
}
~~~

julienrf: I guess there might be a race condition if both tasks try to create the target directory at the same time?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -562,37 +642,51 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
+          dependencyClasspath.value,
           configuration.value,
           (version in webpack).value,
           (version in startWebpackDevServer).value,
           webpackCliVersion.value,
           streams.value
         ),
 
-      crossTarget in npmUpdate := {
+      crossTarget in npmInstall := {
```
julienrf: We should document that `crossTarget in npmUpdate` has to be replaced with `crossTarget in npmInstall` in the changelog. Also, since this is a major breakage we should probably bump the version number to 0.14.0.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
julienrf: E.g.: 

> Use `npmInstall` (or `npmInstallDependencies` if you don’t need .js resources from the classpath)

?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
steinybot: Nope we don't need any particular order, that was just me being silly.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
steinybot: Does having
```
crossTarget in npmInstall := (crossTarget in npmUpdate).value
```
mitigate the effect of this somewhat?

I was hoping previous uses of `crossTarget in npmUpdate` would keep working but now have a deprecation warning.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -100,8 +106,83 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
+      *
+      * The plugin uses different directories according to the configuration (`Compile` or `Test`). Thus,
+      * this setting is scoped by a `Configuration`.
+      *
+      * Typically, if you want to define a task that uses the downloaded NPM packages you should
+      * specify the `Configuration` you are interested in:
+      *
+      * {{{
+      *   myCustomTask := {
+      *     val npmDirectory = (npmInstall in Compile).value
```
dispalt: Is this comment correct?

steinybot: That's a good question. Technically yes but it depends. It should encourage the use of `npmInstallDependencies` instead. I will update it.

steinybot: Oh woops I thought this comment was on the other task but yeah they were both kinda wrong.

julienrf:
Hey, I will probably not be able to review this PR before next week.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
julienrf: Having both this definition and `crossTarget in npmInstall := crossTarget in npmUpdate.value` looks wrong. I think the most migration friendly strategy is to keep `npmUpdate` and `crossTarget in npmUpdate` as they were, and then define `npmInstall := npmUpdate.value` and `crossTarget in npmInstall := (crossTarget in npmUpdate).value`. This way, users that have overridden `npmUpdate` or `crossTarget in npmUpdate` will still have something that works (with deprecation warnings).

steinybot: Your totally right. I think that was what I intended to do I must have just renamed this one by accident. I've moved them together now so it is more obvious what is going on.

julienrf:
Don’t be sorry. I am. I forgot about this PR. I will review it in the next days.

julienrf:
@steinybot Oops, sorry for the delay I forgot to answer you… Yes, I would remove `npmInstall`, what do you think?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -592,14 +640,16 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
-          fullClasspath.value,
+          dependencyClasspath.value,
           configuration.value,
           (version in webpack).value,
           (version in startWebpackDevServer).value,
           webpackCliVersion.value,
           streams.value
         ),
 
+      crossTarget in npmUpdate := (crossTarget in npmUpdate).value,
```
julienrf: We should remove this line now I guess:

```suggestion
```

PR #264, opened 2018-09-15
Add webpackNodeFlags setting

ngbinh:
Allow users to pass https://nodejs.org/api/cli.html to webpack runs

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
@@ -179,7 +179,7 @@ object Webpack {
      entry: BundlerFile.Application,
      targetDir: File,
      extraArgs: Seq[String],
-     nodeFlags: Seq[String],
+     nodeArgs: Seq[String],
      mode: WebpackMode,
      log: Logger
   ): BundlerFile.ApplicationBundle = {
@@ -191,7 +191,7 @@ object Webpack {
 
     log.info("Bundling the application with its NPM dependencies")
     val args = extraArgs ++: Seq("--config", configFile.absolutePath)
-    val stats = Webpack.run(nodeFlags: _*)(args: _*)(targetDir, log)
+    val stats = Webpack.run(nodeArgs: _*)(args: _*)(targetDir, log)
     stats.foreach(_.print(log))
 
     // Attempt to discover the actual name produced by webpack indexing by chunk name and discarding maps
@@ -223,7 +223,7 @@ object Webpack {
     entryPointFile: BundlerFile.EntryPoint,
     libraryModuleName: String,
     extraArgs: Seq[String],
-    nodeFlags: Seq[String],
+    nodeArgs: Seq[String],
     mode: WebpackMode,
     log: Logger
   ): BundlerFile.Library = {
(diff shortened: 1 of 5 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
```diff
@@ -286,13 +288,14 @@ object Webpack {
...
+    * @param nodeFlags node.js cli flags
     * @param args Arguments to pass to the webpack command
     * @param workingDir Working directory in which the Nodejs will be run (where there is the `node_modules` subdirectory)
     * @param log Logger
     */
-  def run(args: String*)(workingDir: File, log: Logger): Option[WebpackStats] = {
+  def run(nodeFlags: String*)(args: String*)(workingDir: File, log: Logger): Option[WebpackStats] = {
     val webpackBin = workingDir / "node_modules" / "webpack" / "bin" / "webpack"
-    val cmd = Seq("node", webpackBin.absolutePath, "--bail", "--profile", "--json") ++ args
+    val cmd = Seq("node") ++ nodeFlags ++ Seq(webpackBin.absolutePath, "--bail", "--profile", "--json") ++ args
```
julienrf: We should then simplify the expression and not create a singleton sequence containing `"node"`.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -348,6 +348,18 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
     )
 
+    /**
+      * node.js cli flags https://nodejs.org/api/cli.html
+      *
+      * Defaults to an empty list.
+      *
+      * @group settings
+      */
+    val webpackNodeFlags = SettingKey[Seq[String]](
```
julienrf: I think it should be named `webpackNodeArgs`, unless `node` uses a special terminology.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -348,6 +348,18 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
...
+      "webpackNodeFlags",
+      "Custom arguments to webpack"
```
julienrf: The description is not correct: it’s arguments to the `node` program when webpack is run.

ngbinh:
update with comments addressed.

ngbinh:
Sorry, I may not be clear on the use case. We need to increase node.js memory on our scala.js bundle build as `weback` would choke from time to time on default memory setting.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -512,7 +524,8 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
     // API user can modify it just once.
     webpackMonitoredDirectories := Seq(),
     (includeFilter in webpackMonitoredFiles) := AllPassFilter,
-    webpackExtraArgs := Seq(),
+    webpackExtraArgs := Seq.empty,
```
cquiroz: So many args :D can't find a good way around though

julienrf:
Thanks @ngbinh! In the future, please avoid changing the code style, it makes it harder to review your work.

Issue #261, opened 2018-08-20
jsdom.createVirtualConsole does not exist in jsdom v12.0.0


thebridsk:
Using scalabundler 0.13.1, scala-js  0.6.24, sbt 1.2.1, scala 2.12.6

```
[info] Writing and bundling the test loader
C:\Users\user\Projects\BridgeScorer\git\bridgescorer\client\target\scalajs-bundler-jsdom\codeWithJSDOMContext.js:13
    virtualConsole: jsdom.createVirtualConsole().sendTo(console),
                          ^
TypeError: jsdom.createVirtualConsole is not a function
```

see https://github.com/scala-js/scala-js/issues/2902

PR #265, opened 2018-09-20
Support jsdom v12.x in JSDOMNodeJSEnv

ngbinh:
Just similar to https://github.com/scala-js/scala-js/issues/3432

should fixes https://github.com/scalacenter/scalajs-bundler/issues/261

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/JSDOMNodeJSEnv.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/JSDOMNodeJSEnv.scala
@@ -46,45 +46,89 @@ class JSDOMNodeJSEnv(
   protected trait AbstractDOMNodeRunner extends AbstractNodeRunner {
 
     protected def codeWithJSDOMContext(): Seq[VirtualJSFile] = {
-      val scriptsJSPaths = getLibJSFiles().map {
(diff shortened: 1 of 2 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/JSDOMNodeJSEnv.scala
```diff
@@ -46,45 +46,89 @@ class JSDOMNodeJSEnv(
   protected trait AbstractDOMNodeRunner extends AbstractNodeRunner {
 
     protected def codeWithJSDOMContext(): Seq[VirtualJSFile] = {
-      val scriptsJSPaths = getLibJSFiles().map {
-        case file: FileVirtualFile => file.path
-        case file                  => libCache.materialize(file).getAbsolutePath
+      val scriptsFiles = (getLibJSFiles() :+ code).map {
```
julienrf: Why do we need this change?

ngbinh: I just copy the changes from https://github.com/scala-js/scala-js/commit/8d1938b6eaa985c933189f0eaf82edee0b3cb897#diff-b097042b8f58e84b044e481289fa803d

PR #266, opened 2018-09-27
Define a custom setting to indicate DOM JsEnv in test

ngbinh:
should allow us to get rid of the already deprecated `requiresDOM` setting in scala.js.

should address #181

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -653,7 +653,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
       npmDevDependencies ++= (npmDevDependencies in Compile).value,
 
       // Default to deprecated requiresDOM to not break old build.
-      requireJsDomEnv := (requiresDOM in Test).value,
+      requireJsDomEnv := requiresDOM.value,
 
       // Override Scala.js setting, which does not support the combination of jsdom and CommonJS module output kind
       loadedTestFrameworks := Def.task {
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -653,7 +653,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
       npmDevDependencies ++= (npmDevDependencies in Compile).value,
 
       // Default to deprecated requiresDOM to not break old build.
-      requireJsDomEnv := (requiresDOM in Compile).value,
+      requireJsDomEnv := requiresDOM.?.value.getOrElse(false),
 
       // Override Scala.js setting, which does not support the combination of jsdom and CommonJS module output kind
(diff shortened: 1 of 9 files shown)


ngbinh:
Thanks. I will update changelog.

cquiroz:
👏👏 this  is great

ngbinh:
Update reference and changelog.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -557,7 +566,9 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
         addPackages(baseDir, installDir, useYarn.value, log, npmExtraArgs.value, yarnExtraArgs.value)(s"jsdom@$jsdomVersion")
       }
       installDir
-    }
+    },
+
+    requireJsDomEnv := false
```
julienrf: What do you think of defaulting to `requiresDOM.value` instead, to not break existing builds?

ngbinh: Make sense! Will do

ngbinh:
updated

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -557,7 +566,10 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
         addPackages(baseDir, installDir, useYarn.value, log, npmExtraArgs.value, yarnExtraArgs.value)(s"jsdom@$jsdomVersion")
       }
       installDir
-    }
+    },
+
+    // Default to deprecated requiresDOM to not break old build.
+    requireJsDomEnv := requiresDOM.value
```
julienrf: This should be in `perConfigSettings` (or even just in `testSettings` if this is only used for testing?)

ngbinh: Make sense. Will try to do that. 

I think the current design may not be as clean as it should. Probably we can just provide our own `jsEnv' and let users switch to it if necessary.

@cquiroz I remember you mentioned before that scala.js bundler relies on some scala.js internal. Is there a chance we can provide a `ScalaJSBundlerDomJsEnv` similar to https://github.com/scala-js/scala-js-dom ?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
```diff
@@ -643,6 +652,9 @@ object ScalaJSBundlerPlugin extends AutoPlugin {
 
       npmDevDependencies ++= (npmDevDependencies in Compile).value,
 
+      // Default to deprecated requiresDOM to not break old build.
+      requireJsDomEnv := (requiresDOM in Compile).value,
```
julienrf: In case `requiresDOM` is not defined we should default to `false`:

```suggestion
      requireJsDomEnv := requiresDOM.?.value.getOrElse(false),
```

ngbinh:
Will work on this one today. Hope to be able to merge

julienrf:
Thanks @ngbinh!

PR #267, opened 2018-09-29
Update scala & sbt to supports JDK9+

exoego:
I noticed that sbt 0.13.16 raises error if JDK9+ is used, so let's update.
Scala 2.11 is also updated to 2.11.12 that supports JDK9+.


exoego:
🤔 
image

```
[info] 'compiler-interface' not yet compiled for Scala 2.10.6. Compiling...
error: error while loading package, Missing dependency 'object java.lang.Object in compiler mirror', required by /home/travis/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.10.6.jar(scala/package.class)
```

I thought sbt 0.13.17 is compiled with [Scala 2.10.7](https://github.com/scala/scala-dev/issues/434) that supports JDK9+, but actually it is not.

julienrf:
Thanks!!

PR #269, opened 2018-10-01
Update docs

ioleo:
In [sbt 1.0.0 release](https://www.scala-sbt.org/1.x/docs/sbt-1.0-Release-Notes.html#sbt+1.0.0) the PathFinder's `***` method has been renamed to `allPaths`.

Using outdated code from the example will yield `error: value *** is not a member of java.io.File`.


julienrf:
Thanks!

PR #271, opened 2018-10-19
Reduce verbosity of test logs

julienrf:
Set the ivyLoggingLevel to Quiet in test builds and test builds builds.

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
@@ -34,7 +34,7 @@ object Commands {
   }
 
   def run(cmd: Seq[String], cwd: File, logger: Logger): Unit = {
-    val toInfoLog = (is: InputStream) => scala.io.Source.fromInputStream(is).getLines.foreach(msg => logger.error(msg))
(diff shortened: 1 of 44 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
```diff
@@ -34,7 +34,7 @@ object Commands {
   }
 
   def run(cmd: Seq[String], cwd: File, logger: Logger): Unit = {
-    val toInfoLog = (is: InputStream) => scala.io.Source.fromInputStream(is).getLines.foreach(msg => logger.error(msg))
+    val toInfoLog = (is: InputStream) => scala.io.Source.fromInputStream(is).getLines.foreach(msg => logger.info(msg))
```
julienrf: @cquiroz I’ve also changed this line to log to the `info` logger rather than the `error` logger. I think that logging to the `error` logger was not intentional.

julienrf:
The purpose of this PR is to reduce the volume of the logs produced by the CI. Unfortunately, I couldn’t find a way to remove the tree of dependencies that is printed by `npm` when we do a `npm install`.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
cquiroz: I think you’re right, seems like a mistake

PR #272, opened 2018-10-31
upgrade to sbt 1.1.0 & play 2.6

mcku:
and passes the test

### DIFF
--- a/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/Loader.scala
+++ b/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/Loader.scala
@@ -1,19 +1,16 @@
 package example
 
-import controllers.{Assets, AssetsConfiguration, DefaultAssetsMetadata}
+import controllers.AssetsComponents
 import play.api.ApplicationLoader.Context
 import play.api.{ApplicationLoader, BuiltInComponentsFromContext}
+import play.filters.HttpFiltersComponents
 import router.Routes
 
 class Loader extends ApplicationLoader {
   def load(context: Context) = new ExampleComponents(context).application
 }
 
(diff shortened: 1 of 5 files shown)


sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/Loader.scala
```diff
@@ -10,7 +10,10 @@ class Loader extends ApplicationLoader {
 }
 
 class ExampleComponents(context: Context) extends BuiltInComponentsFromContext(context) {
+  val httpFilters = Seq()
   val controller = new ExampleController
-  val assets = new Assets(httpErrorHandler)
+  val assetsConfiguration = new AssetsConfiguration()
+  val assetsMetadata = new DefaultAssetsMetadata(assetsConfiguration, environment.resource _, fileMimeTypes)
```
julienrf: There should be `XxxComponents` traits to mix in to our class so that we don’t have to manually instantiate the `AssetsMetadata`, `AssetsConfiguration` and the `HttpFilter` list.

julienrf:
The project you modified fails in the CI, do the tests pass locally?

mcku:
Yes the local tests pass, when run from within `scalajs-bundler/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play`. The log is here https://gist.github.com/mcku/e213153758a9feacae21c3810e0acfdc . 

Looking at the Travis log, it looks like the bundle gets created in `target/scala-2.12/...`. But  `runScripted` looks for `target/scala-2.11`. I don't know where 2.11 originates from. 

Line 2654:
```
Caused by: java.lang.RuntimeException: Test script error: File(s) did not exist: [ /tmp/sbt_b011bfd0/play/client/target/scala-2.11/scalajs-bundler/main/client-fastopt-bundle.js ]
```

sjrd:
Superseded by #352.

PR #281, opened 2019-01-16
Fixes JS files to be included when using library-only bundling mode

MarkusA380:
With the current version the emitted JS files are 
* `yourapp-fastopt-library.js`
* `yourapp-fastopt-loader.js`
* `yourapp-fastopt.js`
(as stated by Julien Richard-Foy on gitter)

### DIFF
--- a/manual/src/ornate/reference.md
+++ b/manual/src/ornate/reference.md
@@ -96,7 +96,7 @@ parameter provided to `BundlingMode.LibraryOnly`.
 It then uses a special "loader" to resolve these dependencies for the actual Scala.js application.
 
 In order to use this mode, instead of including `yourapp-bundle.js` in your page, you will need to include
-`yourapp-library.js`, `yourapp-loader.js`, `yourapp-fastopt.js` or `yourapp-opt.js`. If you're using Play 
+`yourapp-fastopt-library.js`, `yourapp-fastopt-loader.js` and `yourapp-fastopt.js`. If you're using Play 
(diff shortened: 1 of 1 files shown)


julienrf:
I’m wondering why we generate `yourapp-fastopt-library.js` in the first place, instead of `yourapp-library.js`. I see no reason why it could be different between `fastOptJS` and `fullOptJS`. Same for the `loader`.

Maybe @cquiroz or @easel has an idea?

cquiroz:
i'm not sure about the original reason. We can change the name using a webpack config, perhaps that is being observed?

MarkusA380:
I think it should always be `yourapp-library.js`, `yourapp-loader.js` and `yourapp.js`, or when not using the libary mode `yourapp-bundle.js`. Whether you use fastOptJS and fullOptJS should not make a difference in the results, which would simplify any further processing of the scripts (copying, deployment) and make it more beginner friendly overall.

easel:
I think the naming convention was simply trying to mirror the output of the scala-js output file. I don't think theres necessarily any benefit to name-mangling `-loader` and `-library`. 

In full bundling mode perhaps it would be beneficial to keep both versions around since Scala.js does the same?

julienrf:
@MarkusA380 Using different names for the bundles generated by `fullOptJS::webpack` and `fastOptJS::webpack` allows us to cache the result of these tasks. Otherwise we would have to *always* re-evaluate the tasks.

PR #288, opened 2019-03-19
Don't enable ScalaJSWeb

Atry:
`ScalaJSWeb` plug-in will create a source mapping to hash path source file (https://github.com/vmunier/sbt-web-scalajs/blob/1e48af8b505867e12cfcfd212cf4180c87c82ca9/src/main/scala/webscalajs/ScalaJSWeb.scala#L23). Unfortunately, the source file is copied to `target/web/public/main` instead of `client/target/scala-2.12/scalajs-bundler/main`, which is not accessible from "source-map-loader".

As a result, a warning of `NonErrorEmittedError` will be thrown by webpack.

We should disable `ScalaJSWeb` to avoid the warning.

### DIFF
--- a/manual/src/ornate/getting-started.md
+++ b/manual/src/ornate/getting-started.md
@@ -66,7 +66,7 @@ lazy val client = project.enablePlugins(ScalaJSBundlerPlugin)
 
 You also need to setup the `ScalaJSBundlerPlugin` on the Scala.js project, as described in the preceding section, and
 the `sbt-web-scalajs` plugins as described in [their documentation](https://github.com/vmunier/sbt-web-scalajs).
-Note that `sbt-web-scalajs`'s `ScalaJSWeb` plugin must not be enabled, because `ScalaJSWeb` will create a source mapping to source files copied to a hash path, which conflicts with `ScalaJSBundlerPlugin`'s webpack-based source mapping.
+Note that `sbt-web-scalajs`'s `ScalaJSWeb` plugin must not be enabled, because `ScalaJSWeb` will create source mappings to source files copied to a hash path, which conflict with `ScalaJSBundlerPlugin`'s webpack-based source mappings.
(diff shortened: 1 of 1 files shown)


julienrf:
This is not the right way to fix the problem. The `ScalaJSWeb` plugin is used to make the bundle available as an sbt-web asset. If the current implementation doesn’t work correctly we should fix it, but not remove the feature.

Atry:
> The `ScalaJSWeb` plugin is used to make the bundle available as an sbt-web asset. 
Not really. What is used to make `sbt-web` asset is `WebScalaJS`, not `ScalaJSWeb`. The only feature of `ScalaJSWeb` is setup source mapping, which conflicts with `scalajs-bundler`.

julienrf:
Oh, I see. But disabling `ScalaJSWeb` means that source maps won’t work anymore… We should check if it is possible to get everything working.

Atry:
> But disabling ScalaJSWeb means that source maps won’t work anymore… 

In fact the source mapping file is generated properly as long as `ScalaJSWeb` is **disabled**.

julienrf:
Interesting, thanks for the investigation. Could you add a sentence about that (the docs currently say to follow the “normal” installation instructions of scalajs-web)? And also can you fix the other places where the `ScalaJSWeb` plugin is used?

Atry:
Tests are updated in another PR https://github.com/scalacenter/scalajs-bundler/pull/289

julienrf:
Thanks!

Atry:
Another workaround for conflicts between `ScalaJSWeb` and `ScalaJSBundlerPlugin`  is setting `webpackBundlingMode := BundlingMode.LibraryOnly()`, since `ScalaJSBundlerPlugin` does not change Scala.js source mapping in `LibraryOnly`.

Issue #282, opened 2019-01-18
BundlingMode.LibraryOnly pollutes a global namespace


glmars:
`scalajs-bundler` enables `CommonJSModule` module kinde, so `scala-js` doesn't wrap JS output to IIFE wrapper (please, [see here](https://github.com/scala-js/scala-js/blob/0ff1fbe3f7a003482c3b785b6a374202f44f47c1/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/Emitter.scala#L104-L105)) and in this case we have the same problem as described https://github.com/scala-js/scala-js/issues/69 (especially on `fullOptJS `output)

easel:
I'm running into this issue as well. I'm pretty sure when we initially implement LibraryOnly the fastopt / fullopt outputs were wrapped in an IIFE. @julienrf do you happen to know if the requirement on CommonJSModule is new, or if the upstream behavior has changed?

easel:
It appears as if the IIFE on commonjs modules disappeared in scala.js 0.6.16 -- see https://github.com/scala-js/scala-js/issues/2785

As a workaround, you can use Application mode with fullOpt, you'll just lose the properly functioning sourcemaps. So far, I have not been able to get application mode with fullopt to work properly with webpack 4, ymmv.

glmars:
Yes, as workaround we use `BundlingMode.LibraryOnly` only for `fastOptJS`:
```scala
webpackBundlingMode in fastOptJS := BundlingMode.LibraryOnly()
```
Do you think we need a new option for the `scala-js` compiler to preserve IIFE to solve this issue?

easel:
I *think* the correct solution is to consume ES modules instead of CommonJS modules from scala.js, but I could be totally wrong on that. @sjrd would probably know if that's a reasonable direction to pursue.

As a follow up, I did get full application bundling working with Webpack 4 by eliminating webpack minification process. It might be possible to "tune" it instead of turning it off entirely but I haven't pursued that yet.

sjrd:
CommonJS modules do not need to be IIFE'ed. Their specification says that they are semantically scoped in their own scope. If a tool breaks that semantic, it is the tool's fault.

ES modules go even further: they *must not* be IIFE'ed, because `export` and `import` must be at the top-level of the module.

There's nothing Scala.js core can/should do here.

easel:
[...]

PR #292, opened 2019-05-08
 Add support for Scala.js 1.x

julienrf:
Specific source directories have been created for Scala.js 0.6 and
Scala.js 1.x, but most of the code cross-compiles.

Some tests are stuck on Scala.js 0.6 because they use libraryDependencies
that are not yet available for Scala.js 1.x.

Usage of some internal APIs of Scala.js (e.g.
`org.scalajs.core.tools.json`), which are not source compatible between
0.6 and 1.x has been replaced with play-json, which scalajs-bundler
already depended on.

Usage of `JSApp` in tests has been removed.


julienrf:
The following tests are failing with Scala.js 1.0.0-M7:

## sbt-web-scalajs-bundler / play

The build doesn’t even load and reports the following error:

~~~
[info] java.lang.NoClassDefFoundError: org/scalajs/jsdependencies/sbtplugin/JSDependenciesPlugin$
~~~

I don’t understand where we have a dependency on that class.

## sbt-scalajs-bundler / sharedconfig

The project loads and some tests pass, but [this line](https://github.com/scalacenter/scalajs-bundler/blob/scalajs-1.x/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/test#L9) fails. This line loads an HTML page that includes the bundle created by scalajs-bundler. The reported JavaScript error is:

~~~
TypeError: [object Object] is not a function, it is object. (file:/tmp/sbt_c80d120e/sharedconfig/target/scala-2.11/scalajs-bundler/main/sharedconfig-opt-bundle.js#18)
~~~

Of course, line 18 of the file is unreadable because it contains the output of `fullOptJS`. The weird thing is that the same test, with `fastOptJS` does pass (line 4).

Has anyone any clue of what’s happening or want to help to debug?

julienrf:
The following tests are failing with Scala.js 1.0.0-M7:

## sbt-web-scalajs-bundler / play

The build doesn’t even load and reports the following error:

~~~
[info] java.lang.NoClassDefFoundError: org/scalajs/jsdependencies/sbtplugin/JSDependenciesPlugin$
~~~

I don’t understand where we have a dependency on that class.

## sbt-scalajs-bundler / sharedconfig

The project loads and some tests pass, but [this line](https://github.com/scalacenter/scalajs-bundler/blob/scalajs-1.x/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/test#L9) fails. This line loads an HTML page that includes the bundle created by scalajs-bundler. The reported JavaScript error is:

~~~
TypeError: [object Object] is not a function, it is object. (file:/tmp/sbt_c80d120e/sharedconfig/target/scala-2.11/scalajs-bundler/main/sharedconfig-opt-bundle.js#18)
~~~

Of course, line 18 of the file is unreadable because it contains the output of `fullOptJS`. The weird thing is that the same test, with `fastOptJS` does pass (line 4).

Has anyone any clue of what’s happening or want to help to debug?

julienrf:
The following tests are failing with Scala.js 1.0.0-M7:

## sbt-web-scalajs-bundler / play

The build doesn’t even load and reports the following error:

~~~
[info] java.lang.NoClassDefFoundError: org/scalajs/jsdependencies/sbtplugin/JSDependenciesPlugin$
~~~

I don’t understand where we have a dependency on that class.

## sbt-scalajs-bundler / sharedconfig

The project loads and some tests pass, but [this line](https://github.com/scalacenter/scalajs-bundler/blob/scalajs-1.x/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/test#L9) fails. This line loads an HTML page that includes the bundle created by scalajs-bundler. The reported JavaScript error is:

~~~
TypeError: [object Object] is not a function, it is object. (file:/tmp/sbt_c80d120e/sharedconfig/target/scala-2.11/scalajs-bundler/main/sharedconfig-opt-bundle.js#18)
~~~

Of course, line 18 of the file is unreadable because it contains the output of `fullOptJS`. The weird thing is that the same test, with `fastOptJS` does pass (line 4).

Has anyone any clue of what’s happening or want to help to debug?

julienrf:
The following tests are failing with Scala.js 1.0.0-M7:

**`sbt-web-scalajs-bundler / play`**

The build doesn’t even load and reports the following error:

~~~
[info] java.lang.NoClassDefFoundError: org/scalajs/jsdependencies/sbtplugin/JSDependenciesPlugin$
~~~

I don’t understand where we have a dependency on that class.

**`sbt-scalajs-bundler / sharedconfig`**

The project loads and some tests pass, but [this line](https://github.com/scalacenter/scalajs-bundler/blob/scalajs-1.x/sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sharedconfig/test#L9) fails. This line loads an HTML page that includes the bundle created by scalajs-bundler. The reported JavaScript error is:

~~~
TypeError: [object Object] is not a function, it is object. (file:/tmp/sbt_c80d120e/sharedconfig/target/scala-2.11/scalajs-bundler/main/sharedconfig-opt-bundle.js#18)
~~~

Of course, line 18 of the file is unreadable because it contains the output of `fullOptJS`. The weird thing is that the same test, with `fastOptJS` does pass (line 4).

Has anyone any clue of what’s happening or want to help to debug?

sjrd:
For the sbt-web-scalajs-bundler error, that's probably because it unconditionally depends on `"com.vmunier" % "sbt-web-scalajs" % "1.0.8-0.6"` which is clearly for Scala.js 0.6.x. I guess the dependency should be adapted according to `scalaJSVersion`.

julienrf:
Indeed, then we have another issue: sbt-web-scalajs depends on 1.0.0-M3, which is binary incompatible with 1.0.0-M7. I will make a PR on sbt-web-scalajs to update to 1.0.0-M7.

ngbinh:
>Of course, line 18 of the file is unreadable because it contains the output of fullOptJS. The weird thing is that the same test, with fastOptJS does pass (line 4).

I am no expert but could it be the new GCC and Common Module?

julienrf:
I think I’ve found the problem with the remaining failing tests. Unless I’m wrong, the problem was present before as well, but just didn’t show up. The `fullOptJS` output uses a name that is also used by a library (leaflet) that the program depends on. So, they just clash. Is there a way to tell the Scala.js optimizer to not use certain reserved identifiers?

sjrd:
> Is there a way to tell the Scala.js optimizer to not use certain reserved identifiers?

No, that's not possible, because Scala.js shouldn't have to ignore certain identifiers. CommonJS modules have dedicated scopes per spec. The bundler shouldn't mangle the scopes of different modules together.

julienrf:
I’ve pushed a commit that disables the Scala.js optimizer on that test, because I think that’s the only solution for now. In the long run, I would like to remove the `LibraryAndApplication` and `LibraryOnly` bundling modes, because they are just hacks that try make it possible to execute CommonJS modules in the browser, but this example shows that they are not robust options.

glmars:
@julienrf 
> Unless I’m wrong, the problem was present before as well, but just didn’t show up

No. Some time ago Scala.js team removed IIFE wrapper around a module output and the problem with `LibraryOnly` mode was introduced. See https://github.com/scalacenter/scalajs-bundler/issues/282 for details.

The `LibraryOnly` mode is important feature of `scalajs-bundler`, without it the compilation speed is unacceptable. We should found a way to preserve it!

dispalt:
I agree with keeping LibraryOnly and don't mean to pile on, it makes compilation fast for dev, which is basically the only way the Compile->Run loop is tolerable.

julienrf:
> The `LibraryOnly` mode is important feature of `scalajs-bundler`, without it the compilation speed is unacceptable. We should found a way to preserve it!

Even with the latest webpack, is it still too slow?

ramnivas:
> Even with the latest webpack, is it still too slow?

Yes, it is quite slow even with the latest webpack. I don't think webpack can do anything better since the core issue is huge single fastOptJs file (about 27 MB in our case), so webpack has to at least parse it every time and that is going to take significant time. 

Until Scala.js produces multiple small js files with only a few changing in each incremental compilation (which would be a wonderful feature by itself, anyway), removing `LibraryOnly` mode will be very detrimental to developer productivity.

glmars:
> Even with the latest webpack, is it still too slow?

The latest webpack still parses a huge Scala.js output. So the speed is the same. Thanks @easel for introducing the `LibraryOnly` mode, it is the only option to  develop a real Scala.js application comfortably.

exoego:
Our team cannot use LibraryOnly mode for some reasons, so now use default mode with a WebPack configuration like

```js
const webpack = require('webpack');
const merge = require("webpack-merge");
const TerserPlugin = require("terser-webpack-plugin");
const scalajsBundlerConfig = require("./scalajs.webpack.config.js");

module.exports = merge(scalajsBundlerConfig, {
    target: 'node',
    output: {
        libraryTarget: "commonjs2"
    },
    optimization: {
        minimizer: [
            new TerserPlugin({
                terserOptions: {
                    // Disable since Scala.js is already compressed by GCC
                    compress: false,
                    // mangling is also disabled due to trade off between optimization-time and space-efficiency
                    mangle: false,
                    output: {
                        comments: false,
                    },
                },
            }),
        ],
    },
});
```

which reduces optJS time reasonable amount.

glmars:
@exoego Did you write this about a `fastOptJS` task? As I know there are all optimizations already disabled (GCC isn't used) in `fastOptJS`.

exoego:
@glmars The above was meant for fullOptJS

glmars:
Ah, ok, but a speed of `fullOptJS` isn't so important for developers. The `LibraryOnly` mode is especially significant in `fastOptJS` to improve a Change->Compile->Run->Change ... cycle

sbt-scalajs-bundler/src/main/scala-sjs-1.x/scalajsbundler/sbtplugin/Settings.scala
```diff
@@ -22,97 +22,69 @@ private[sbtplugin] object Settings {
...
-      val (env, input) = {
-        Def.taskDyn[(JSEnv, org.scalajs.jsenv.Input)] {
+  val jsEnvTestSettings =
+    Seq(
+      jsEnv := Def.taskDyn[JSEnv] {
+        if (requireJsDomEnv.value) Def.task {
+          val jsdomDir = installJsdom.value
+          new JSDOMNodeJSEnv(JSDOMNodeJSEnv.Config(jsdomDir))
+        } else Def.task {
+          jsEnv.value
```
sjrd: I'm pretty sure this will turn into a cyclic error. A `taskDyn` unfortunately cannot conditionally depend on its previous value. But perhaps it can work with the following rearrangement:
```scala
jsEnv := {
  val defaultJSEnv = jsEnv.value
  Def.taskDyn[JSEnv] {
    if (requireJsDomEnv.value) Def.task {
      val jsdomDir = installJsdom.value
      new JSDOMNodeJSEnv(JSDOMNodeJSEnv.Config(jsdomDir))
    } else Def.task {
      defaultJSEnv
    }
  }
}
```

sbt-scalajs-bundler/src/main/scala-sjs-1.x/scalajsbundler/sbtplugin/Settings.scala
```diff
@@ -22,97 +22,69 @@ private[sbtplugin] object Settings {
...
-        case (tf, Some(adapter)) => (tf, adapter)
-      }.toMap: Map[TestFramework, Framework]
-    }.dependsOn(npmUpdate).value
+              Set.empty
+            }
+          writeTestBundleFunction(Set(sjsOutput))
+          val file = new FileVirtualBinaryFile(bundle)
+          org.scalajs.jsenv.Input.ScriptsToLoad(List(file))
+        } else Def.task {
+          jsEnvInput.value
```
sjrd: There will be a similar problem here, I believe.

julienrf:
Thanks to all for your feedback on the usage of the `Library*` bundling modes. You convinced me that we should keep it for now ;)

I’ve rebased the PR, I think it is ready to be merged.

cquiroz:
Thanks a lot @julienrf. this is a key component for the 1.x ecosystem

oyvindberg:
One more huge thanks for this, now Scala.js milestones are finally useful for me! :)

PR #298, opened 2019-05-23
Copy nested directory contents from jsSourceDirectories

### LINKED ISSUE #296: [0.15.0] Directory within a `jsSourceDirectories` gives error
Using `0.15.0-0.6`, I am trying to make use of the key `jsSourceDirectories`. I have the following hierarchy of some resource files
```
resources
|____types
        |____index.d.ts
```
If I set the `jsSourceDirectories` to point to the root `resources` directory I get the following exception `Source file 'resources/types' is a directory`. However, the functionality should have been recursive and resolve all the files or the sub-directories can be copied as is. Is this a bug?

PuneetPatwari:
Fixes #296


julienrf:
Thanks for the quick PR accompanied of a test!

PR #304, opened 2019-05-29
CI ONLY Intercept LinkingUnit in fastOptJS linker

gzm0:
This allows us to only link once, instead of running linking twice,
once to actually link and once to retrieve a LinkingUnit.


gzm0:
I'm trying to figure out what broke here. CI says the following tests fail:

- sbt-scalajs-bundler / additonalNpmConfig
- sbt-scalajs-bundler / library
- sbt-scalajs-bundler / static

Is there a way to run them selectively?

julienrf:
Yes, see the “run a single test” section in the [contributing](https://github.com/scalacenter/scalajs-bundler/blob/master/CONTRIBUTING.md#run-the-tests) file.

julienrf:
I believe this PR is not really relevant anymore because of #325? Feel free to re-open it.

PR #305, opened 2019-05-29
Log warning during fullOptJS when using a Library bundling mode

### LINKED ISSUE #301: A warning should be printed when LibraryOnly is used with fullOptJS
The documentation [clearly says](https://scalacenter.github.io/scalajs-bundler/reference.html#bundling-mode-library-only) that the `LibraryOnly` bundling mode should not be used with `fullOptJS`. We should also print a warning when this case is detected.

andrewrembrandt:
Fixes #301 (also required for LibraryAndApplication)


andrewrembrandt:
You're welcome @julienrf - hopefully I get some time later this year to 
help out scalajs wise - do let me know if there's something appropriate 
in the mean time however.

PR #307, opened 2019-06-06
Add the generated config file, so it gets picked up by the cache

dispalt:
This is a simple attempt to solve a double run issue I noticed with the FileFunction.cached, I could put this in the `WebpackTasks.bundle` function after the .cached call, but that seemed less clean.


dispalt:
I actually don't think this fixes it, nor is it the right way to fix it.  

Problem statement:
- with a new clean repo, if you run fullOptJS::webpack, you could run it twice in a row before the caching kicks in.  This hurts on CI/CD specifically because I have tasks that analyze the result and do something it.
- I debugged it to the only file changed in the FileFunctions.cache function in the second run is scalajs.webpack.config.js.
- I think it has to do with the result of the cache function Set[File] => Set[File] that particular files changed, and the generated file isn't present, so it reruns.

There's probably something I am missing.

Issue #293, opened 2019-05-17
Support ModuleKind.ESModule


kpritam:
Latest Scala.js supports ModuleKind.ESModule, this option can not be used with ScalaJsBundler, we see this error:

```txt
java.lang.RuntimeException: scalaJSModuleKind must be set to ModuleKind.CommonJSModule in projects where ScalaJSBundler plugin is enabled
```

julienrf:
Indeed, I think checking that `scalaJSModuleKind` has been set to something **different than `ModuleKind.NoModule`** should work. Ideally, we could also check that the webpack version used does support ES modules.

vhiairrassary:
As a POC, when deactivating `moduleKind` check:
```diff
diff --git a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
index ace7f71..a8f6a93 100644
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/ScalaJSBundlerPlugin.scala
@@ -569,7 +569,7 @@ object ScalaJSBundlerPlugin extends AutoPlugin {

     ensureModuleKindIsCommonJSModule := {
       if (scalaJSLinkerConfig.value.moduleKind == ModuleKind.CommonJSModule) true
-      else sys.error(s"scalaJSModuleKind must be set to ModuleKind.CommonJSModule in projects where ScalaJSBundler plugin is enabled")
+      else true
     },

     webpackBundlingMode := BundlingMode.Default,
```

with default `webpackBundlingMode` and
```scala
webpackCliVersion := "3.3.6"
version in webpack := "4.39.1"
version in startWebpackDevServer := "3.8.0"

scalaJSLinkerConfig ~= { _.withModuleKind(ModuleKind.ESModule) }
```
it just works with Scala.js `0.6.28`, Chrome 76 by loading `[name]-fastopt-bundle.js`. Are there other areas to check before starting a complete pull request?

julienrf:
Awesome, thanks for investigating!

> Are there other areas to check

Maybe running tests, and running tests that use the DOM.

KadekM:
Any news on this?

[...]

PR #312, opened 2019-08-19
[WIP] Add ES modules support

vhiairrassary:
Fixes https://github.com/scalacenter/scalajs-bundler/issues/293

Allow to use ES modules as well as CommonJS module.

**WIP**: `BundlingMode.LibraryOnly()` does not work yet as with:
```html
<script src="[name]-fastopt-library.js"></script>
<script src="[name]-fastopt-loader.js"></script>
<script src="[name]-fastopt.js"></script>
```
the browser will not recognize the `import` syntax in the last script.

Adding the attribute `type="module"` is not enough: if `[name]-fastopt.js` contains something like `import * as $i_jquery from "jquery";` then the following error is thrown:
```
Uncaught TypeError: Failed to resolve module specifier "jquery". Relative references must start with either "/", "./", or "../".
```

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala-sjs-0.6/scalajsbundler/sbtplugin/Settings.scala
+++ b/sbt-scalajs-bundler/src/main/scala-sjs-0.6/scalajsbundler/sbtplugin/Settings.scala
@@ -11,7 +11,7 @@ import org.scalajs.sbtplugin.ScalaJSPluginInternal.{scalaJSEnsureUnforked, scala
 import sbt.Keys.{loadedTestFrameworks, streams, testFrameworks, version}
 import sbt._
 import scalajsbundler.sbtplugin.ScalaJSBundlerPlugin.autoImport.{installJsdom, npmUpdate, requireJsDomEnv, webpack, webpackConfigFile, webpackNodeArgs, webpackResources}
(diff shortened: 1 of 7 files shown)


julienrf:
> `BundlingMode.LibraryOnly()` does not work

I’m not sure what we can do about that. By design, the `LibraryOnly` mode can only work with a module kind that is understood by web browsers. We could just document that the `LibraryOnly` mode only works with `CommonJSModule`.

a1russell:
Any update on this?

jeroentervoorde:
I updated the PR against current master. It seems to work fine except for this warning when trying to configure ES modules in fullOptJS (scalaJSLinkerConfig in fullOptJS ~= (_.withModuleKind(ModuleKind.ESModule)) ):

"The module kind in `scalaJSLinkerConfig in (learnreact, compile, fullOptJS)` is different than the one `in `(learnreact, compile)`. Some things will go wrong."

Not sure what to do about that. We should be able to configure ES modules in fullOptJS only i guess as it does not work will BundlingMode.LibraryOnly()

My reason to enable this is fullOptJS is the smaller bundler size. It generates a 25% smaller output file. Probably because of the better DCE for ES modules in webpack.

Updated PR.
https://github.com/scalacenter/scalajs-bundler/pull/336

vhiairrassary:
Closing in favor of https://github.com/scalacenter/scalajs-bundler/pull/336

PR #313, opened 2019-08-30
[WIP] Custom environment variables to node.js when running webpack tasks

vhiairrassary:
### WIP ⚠️:
- Only one function of `util.Commands.scala`  has been updated to accept extra env. vars. Should I update the others too?

If adding this new option seems ok to the maintainers then I can add a test.

### Usage example:
```scala
webpackNodeEnvVars in fullOptJS := Map(
  "NODE_ENV" -> "production"
)
```

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala-sjs-0.6/scalajsbundler/sbtplugin/Settings.scala
+++ b/sbt-scalajs-bundler/src/main/scala-sjs-0.6/scalajsbundler/sbtplugin/Settings.scala
@@ -10,7 +10,7 @@ import org.scalajs.sbtplugin.ScalaJSPlugin.AutoImport.{fastOptJS, jsEnv, loadedJ
 import org.scalajs.sbtplugin.ScalaJSPluginInternal.{scalaJSEnsureUnforked, scalaJSModuleIdentifier}
 import sbt.Keys.{loadedTestFrameworks, streams, testFrameworks, version}
 import sbt._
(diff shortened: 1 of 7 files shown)


julienrf:
Thanks @vhiairrassary for the PR!

We already have the `npmExtraArgs` setting, could it be used to pass environment variables as well?

vhiairrassary:
@julienrf From my understanding webpack [is not launched using `npm`](https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala#L299). Did I miss something?

julienrf:
Indeed, `webpackNodeArgs` is the correct setting name. (We have so many `*Args` settings…)

julienrf:
But, according to [the documentation](https://nodejs.org/api/cli.html) it seems that env variables don’t have arguments equivalent, so we must provide a way to pass env variables, right?

vhiairrassary:
I did not use the name `webpackNodeArgs` as there is no way to set env. var. for a Node.js script using command line options.

So I created `webpackNodeEnvVars` which is different (as it set the launched process environnent) but I also understand that adding an other similar setting is not ideal 😔

julienrf:
Thanks, I now completely understand the issue addressed by this PR, and I agree that we should address this issue.

I’m a bit worried by all these `*Args` keys defined by the `ScalaJSBundlerPlugin`, but I believe the problem is more general, I’ve written my thoughts about it in #314. In the meantime, I’m happy to merge your PR after all the places where we invoke `node` are updated.

vhiairrassary:
Hum, after having read https://github.com/scalacenter/scalajs-bundler/issues/314#issuecomment-527176298, I understand why these variables (such as `webpackNodeEnvVars`) may be out of scope for this project. Now I am sure it is a good idea to merge that PR 😄

PR #316, opened 2019-12-03
Upgrade to Scala.js 1.0.0-RC1.

sjrd:
This includes the minimal changes required to upgrade the Scala.js 1.x support to 1.0.0-RC1, dropping 1.0.0-M7. The main limitation is that we make sbt-scalajs-bundler directly depend on scalajs-linker, instead of loading scalajs-linker by reflection like sbt-scalajs does. This preserves the status quo of 1.0.0-M8 instead of benefitting from the improvements of 1.0.0-RC1 in that area, so it could use improvements in the future but is correct for the moment.


julienrf:
It looks like some dependencies (eg scalatest) used in the tests need to be bumped.

sjrd:
It looks like we'll need ScalaTest to publish for 1.0.0-RC1 first. Filed as https://github.com/scalatest/scalatest/issues/1729.

sjrd:
I made it work by converting ScalaTest tests to JUnit.

@julienrf Do you have a requirement that all commits must pass tests? If yes I'll invert the two commits of this PR, since the first one won't pass tests as is.

julienrf:
> Do you have a requirement that all commits must pass tests?

No.

sjrd:
ScalaTest 3.1.0 was released for Scala.js 1.0.0-RC1, so I removed the second commit from this PR. We might still want to consider it in a separate PR for the future (and particularly the eventual release of Scala.js 1.0.0 final) but it can be reviewed later in a separate PR.

julienrf:
> We might still want to consider it in a separate PR for the future

Yes, that’d be simpler for us. You can create a new PR with it, if you want (maybe after we merge this one, it will be simpler).

sjrd:
> maybe after we merge this one, it will be simpler

Yes, it would definitely be simpler, so I'll wait until this PR is merged to resubmit the change to JUnit.

build.sbt
```diff
@@ -16,6 +16,12 @@ val `sbt-scalajs-bundler` =
       description := "Module bundler for Scala.js projects",
       libraryDependencies += "com.typesafe.play" %% "play-json" % "2.6.7",
       addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion),
+      libraryDependencies ++= {
+        if (isScalaJS1x)
+          List("org.scala-js" %% "scalajs-linker" % scalaJSVersion)
```
julienrf: So, a possible drawback could be that this could conflict with other build dependencies? Do you think we could find a way to not have to implement this proxy over the Scala.js linker?

sjrd: It is a possible drawback, but that drawback already exists today with sbt-scalajs 0.6.x and with scalajs-bundler for 0.6.x, because before sbt-scalajs depended itself on scalajs-linker.

We should find a way to improve this but it really should be a different PR.

PR #317, opened 2019-12-09
Replace ScalaTest by JUnit in all Scala.js test projects.

sjrd:
This frees us from having to wait for ScalaTest to publish new releases compatible with Scala.js 1.x. Since the JUnit is provided by the core Scala.js team, it is guaranteed to be readily available. And given the trivial test suites used in the testing projects, using ScalaTest is not justified.


cquiroz:
Great idea!

PR #318, opened 2019-12-16
[WIP] Upgrade to Scala.js 1.0.0-RC2

FabioPinheiro:
Support of Scala.js 1.0.0-RC2 instead of 1.0.0-RC1


FabioPinheiro:
Closeing this PR
Note: sbt-scalajs-bundler version 0.16.0 does not work with scalajs 1.0.0-RC2 because is complied with scalajs 1.0.0-RC1
As workaround by @sjrd is to use the scalajs-linker:
`libraryDependencies += "org.scala-js" %% "scalajs-linker" % "1.0.0-RC2"`
See https://github.com/scalacenter/scalajs-bundler/pull/319#issuecomment-569463765

PR #324, opened 2020-02-17
Upgrade to Scala.js 1.0.0, scalajs-dom 1.0.0 and sbt-web-scalajs 1.0.11.

sjrd:


.travis.yml
```diff
@@ -13,8 +13,8 @@ env:
 jdk:
   - openjdk11
 before_install:
-- nvm install 6
-- nvm use 6
+- nvm install 12
```
julienrf: Should we also add a note about this requirement in the docs?

sjrd: I did not find a place in the docs that currently mentioned what version of Node.js is required. I'm happy to add a note if you can point me to an appropriate place.

julienrf: I would add it in the note [here](https://github.com/scalacenter/scalajs-bundler/blob/master/manual/src/ornate/getting-started.md#basic-setup)

sjrd: Hum, I don't think that's appropriate. Those are *user* documentation, but a recent Node.js is only required for some *tests* of sbt-scalajs-bundler, so it is only require when *developing* sbt-scalajs-bundler.

julienrf: OK, I didn’t notice that this requirement applied only to run sbt-scalajs-bundler’s test. Is there a way to check that users of sbt-scalajs-bundler don’t need a more recent version of npm?

sjrd: Not that I can think of, short of having a very complicated CI.

PR #325, opened 2020-02-18
With Scala.js 1.x, do not relink to compute the imported modules.

sjrd:


sjrd:
The CI failure on the first commit is spurious.

PR #328, opened 2020-02-20
Upgrade to sbt 1.3.8 and use Scala.js 1.x by default.

sjrd:
Upgrading to sbt 1.x required upgrading all sbt plugins.

Using Scala.js 1.x by default provides a better experience in Metals out of the box, because of the `scalajs-bundler-linker` project which cannot resolve with Scala.js 0.6.x.

The sbt plugin is still compiled against sbt 1.2.8, so that users of sbt-scalajs-bundler are not forced to upgrade to sbt 1.3.x themselves.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,9 +1,6 @@
-import sbtunidoc.Plugin.ScalaUnidoc
-import sbtunidoc.Plugin.UnidocKeys.unidoc
-
 val runScripted = taskKey[Unit]("Run supported sbt scripted tests")
 
-val scalaJSVersion = sys.env.getOrElse("SCALAJS_VERSION", "0.6.31")
+val scalaJSVersion = sys.env.getOrElse("SCALAJS_VERSION", "1.0.0")
 val isScalaJS1x = scalaJSVersion.startsWith("1.")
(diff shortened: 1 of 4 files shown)


sjrd:
I locally tested that `manual/ornate` and `apiDoc/unidoc` work and produce appropriate results.

I'm not sure how to test the consequences of upgrading sbt-ghpages and sbt-pgp.

build.sbt
```diff
@@ -17,7 +14,7 @@ lazy val `scalajs-bundler-linker` =
 
 val `sbt-scalajs-bundler` =
   project.in(file("sbt-scalajs-bundler"))
-    .enablePlugins(BuildInfoPlugin)
+    .enablePlugins(SbtPlugin, BuildInfoPlugin)
```
julienrf: What is the difference with setting `sbtPlugin := true`?

sjrd: It also enables `scripted`. In sbt 1.x you have to use that to use `scripted` tests.

sjrd:
Btw, after this PR I think we're completely ready to release a new version of scalajs-bundler that works well with Scala.js 1.0.0+ out of the box.

PR #334, opened 2020-03-21
Load a newer version of the linker to match the version of sbt-scalajs.

sjrd:
With Scala.js 1.x, scalajs-bundler loads a custom linker with a custom backend in `Global / scalaJSLinkerImpl / fullClasspath`. Since the artifact with the custom linker depends on `scalajs-linker`, we also receive that one. However, that always loads the version of `scalajs-linker` against which scalajs-bundler was built. If we combine it with a newer version of `sbt-scalajs`, we still loaded the old version of `scalajs-linker`. This can prevent scalajs-bundler from being used with non forward compatible versions of Scala.js, or simply to miss out on bug fixes and improvements from upstream.

This commit solves this issue by adding an explicit dependency on scalajs-linker with the same version as that of sbt-scalajs (which we find in `scalaJSVersion`). As long as that version of the linker is backward binary compatible with our usage in the custom linker, this will allow scalajs-bundler to be used with more recent versions of sbt-scalajs.

We test this with a test that specifically bumps sbt-scalajs to 1.0.1 and contains a test for a bugfix introduced in the upstream linker version 1.0.1.


sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/newer-linker_sjs-1/project/newer-scala-js.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
```
julienrf: IIUC we should not update the Scala.js dependency of scalajs-bundler otherwise this test won’t test anything, right?

sjrd: That's right.

cquiroz:
Could you please make a release containing this fix so we could use scalajs-bundler with scala.js 1.1.0?

julienrf:
Indeed, we should do a new release. Would you be interested in taking care of it?

cquiroz:
Sure, I'm guessing just tagging will do it? I'll merge #339 before

PR #336, opened 2020-04-06
Vhiairrassary/es modules support

jeroentervoorde:


sjrd:
The warning is legitimate, and you should heed it. You'll run into all sorts of issues if you set the module kind `in fullOptJS` differently from not `in fullOptJS` (same with `in fastOptJS`). sbt-scalajs assumes that they are not overridden per task, which is why it emits that warning.

jeroentervoorde:
Thanks. I already assumed you put that warning there for a reason :) 
This workaround works for me:
```
sbt "set scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.ESModule))" clean fullOptJS::webpack
```

If i update the documentation would you / @julienrf be willing to merge this?

dispalt:
I'd like to see this as well, we get a major size reduction with es6 too

dispalt:
Actually that is incorrect, I don't get any savings. I must have been mixing something up in my head with 0.6... Sorry for the noise, but fullOpt shouldn't get any savings because of this. https://github.com/scala-js/scala-js/issues/3893

nightscape:
Any news on this?

PR #340, opened 2020-05-16
Prepare next release

cquiroz:
This should get scalajs-bundler ready for the next release.
I tested it with a local publish and a project set to use scala.js 1.1.0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 val runScripted = taskKey[Unit]("Run supported sbt scripted tests")
 
-val scalaJSVersion = sys.env.getOrElse("SCALAJS_VERSION", "1.1.0")
+val scalaJSVersion = sys.env.getOrElse("SCALAJS_VERSION", "1.0.0")
 val isScalaJS1x = scalaJSVersion.startsWith("1.")
 val scalaJSSourceDirectorySuffix = if (isScalaJS1x) "sjs-1.x" else "sjs-0.6"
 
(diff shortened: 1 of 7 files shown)


sjrd:
Oh, no, we really shouldn't do that. That's not necessary. We can already use the `master` of scalajs-bundler with Scala.js 1.1.0, thanks to #334. There is a test that makes sure it works with 1.0.1; at most we could add a test to make sure it also works with 1.1.0. But if we *build* scalajs-bundler with 1.1.0, it won't be usable with 1.0.x anymore.

In general, we should never upgrade the version of Scala.js used to build scalajs-bundler, unless it unlocks something new in scalajs-bundler that we couldn't do otherwise.

cquiroz:
@sjrd Thanks for the info. I reverted the upgrade to 1.1.0 though I set it to 1.1.0 in one of the tests as a sanity check

build.sbt
```diff
@@ -1,14 +1,14 @@
 val runScripted = taskKey[Unit]("Run supported sbt scripted tests")
 
-val scalaJSVersion = sys.env.getOrElse("SCALAJS_VERSION", "1.0.0")
+val scalaJSVersion = sys.env.getOrElse("SCALAJS_VERSION", "1.1.0")
```
sjrd, line 3: This should stay 1.0.0. Same for the versions in .travis.yml. Otherwise effectively we're still building against 1.1.0.

cquiroz: Could it be 1.0.1?

cquiroz:
Thanks, I'll cut a release

PR #343, opened 2020-05-18
Drop support for sbt 0.13.x.

sjrd:
This doesn't work. We need to update the Play test in sbt-web-scalajs-bundler to at least Play 2.6.x, which has sbt 1.x support. I try in the second commit, but I can't make the Play test to work. I'm just going to leave this here for someone else to pick up.


vmunier:
Just opened a PR to upgrade the Play sbt-test to the latest dependency versions https://github.com/scalacenter/scalajs-bundler/pull/352

sjrd:
Rebased, now that #352 has been merged.

build.sbt
```diff
@@ -152,14 +156,8 @@ lazy val commonSettings = List(
     "-Dsbt.execute.extrachecks=true" // Avoid any deadlocks.
   ),
   scriptedBufferLog := false,
-  crossSbtVersions := List("0.13.17", "1.2.8"),
+  crossSbtVersions := List("1.2.8"),
   sbtVersion in pluginCrossBuild := "1.2.8",
```
julienrf, line 160: Should we use 1.3.0 everywhere, for simplicity?

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sbt-1.0.x/build.sbt
```diff
@@ -1,6 +0,0 @@
-enablePlugins(ScalaJSBundlerPlugin, ScalaJSJUnitPlugin)
-scalaVersion := "2.12.3"
-scalaJSUseMainModuleInitializer := true
-npmDependencies in Compile += "left-pad" -> "1.1.3"
-
-ivyLoggingLevel in ThisBuild := UpdateLogging.Quiet
```
julienrf, line 6: Shouldn’t we keep this test?

build.sbt
sjrd: If we do that, then users of sbt-scalajs-bundler would be forced to upgrade to sbt 1.3.x. Is that something we want to do?

Even if yes, I believe that should be in a separate PR.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sbt-1.0.x/build.sbt
sjrd: It doesn't seem to test anything that isn't tested several times over in the other tests, so I thought it might as well be removed.

build.sbt
julienrf: No, ok for staying on 1.2

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/sbt-1.0.x/build.sbt
julienrf: OK, I didn’t double check, thanks!

PR #349, opened 2020-05-21
Add additionalNpmConfig and targetDir to the hash

MasseGuillaume:


cquiroz:
What do you get with this change? Will it make it re run if aditionalNpmConfig or targetDir change?

PR #358, opened 2020-08-14
Update static test

cquiroz:
Upgrade the `static` test. Updates scala to 2.13 and removes uglify since it is not needed with newer webpack versions


cquiroz:
Actually this seems to be falling with scala.js 0.6. I wonder if I should make this test 1.x only

PR #361, opened 2020-09-14
Remove obsolete warning regarding scalajs-scripts usage

mn98:
WebScalaJSBundlerPlugin works just fine with scalajs-scripts since [this update](https://github.com/vmunier/scalajs-scripts/blob/2a37ef7cde4ea6ce0f40e0d7318ea5f218f765f2/src/main/twirl/scalajs/selectScript.scala.html#L9).

### DIFF
--- a/manual/src/ornate/reference.md
+++ b/manual/src/ornate/reference.md
@@ -187,12 +187,6 @@ where you used to enable `WebScalaJS`).
 The plugin tunes the `scalaJSPipeline` to use the bundles produced by webpack rather than the direct
 output of the Scala.js compilation.
 
-> {.note}
-> The [scalajs-scripts](https://github.com/vmunier/scalajs-scripts) will **not** work anymore because they
-> try to include the output of Scala.js instead of the output of the bundling process. You can see
-> [here](https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/ExampleController.scala#L25-L30)
-> how to include the correct .js file.
-
 ### Importing Assets from NPM Packages {#npm-assets}
 
 Some NPM packages also contain static assets (e.g. fonts, stylesheets, images, etc.). You can make them available
--- a/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/ExampleController.scala
+++ b/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/ExampleController.scala
@@ -6,7 +6,6 @@ import play.twirl.api.StringInterpolation
 class ExampleController(cc: ControllerComponents) extends AbstractController(cc) {
 
   val index = {
-    val scriptUrl = bundleUrl("client")
     val result =
       Ok(
         html"""<!doctype html>
@@ -14,7 +13,7 @@ class ExampleController(cc: ControllerComponents) extends AbstractController(cc)
           <head></head>
           <body>
             <div>App is not loaded.</div>
-            <script src="$scriptUrl"></script>
+            @scalajs.html.scripts("client", routes.Assets.versioned(_).toString, name => getClass.getResource(s"/public/$$name") != null)
           </body>
         </html>
       """
@@ -22,10 +21,4 @@ class ExampleController(cc: ControllerComponents) extends AbstractController(cc)
     Action(result)
   }
 
-  def bundleUrl(projectName: String): Option[String] = {
-    val name = projectName.toLowerCase
-    Seq(s"$name-opt-bundle.js", s"$name-fastopt-bundle.js")
-      .find(name => getClass.getResource(s"/public/$name") != null)
-      .map(controllers.routes.Assets.versioned(_).url)
-  }
 }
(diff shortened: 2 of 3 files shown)


sjrd (approved):

manual/src/ornate/reference.md
```diff
@@ -187,12 +187,6 @@ where you used to enable `WebScalaJS`).
 The plugin tunes the `scalaJSPipeline` to use the bundles produced by webpack rather than the direct
 output of the Scala.js compilation.
 
-> {.note}
-> The [scalajs-scripts](https://github.com/vmunier/scalajs-scripts) will **not** work anymore because they
-> try to include the output of Scala.js instead of the output of the bundling process. You can see
-> [here](https://github.com/scalacenter/scalajs-bundler/blob/master/sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/ExampleController.scala#L25-L30)
```
julienrf, line 193: Thanks for fixing the instructions. If you have time, it would be even better to update that test to use the `scalajs-scripts` instead of our _ad-hoc_ technique.

mn98: Ok, I've pushed an update to the test, it appears to run fine locally for me.

sbt-web-scalajs-bundler/src/sbt-test/sbt-web-scalajs-bundler/play/server/src/main/scala/example/ExampleController.scala
```diff
@@ -6,26 +6,19 @@ import play.twirl.api.StringInterpolation
...
-    val scriptUrl = bundleUrl("client")
     val result =
       Ok(
         html"""<!doctype html>
         <html>
           <head></head>
           <body>
             <div>App is not loaded.</div>
-            <script src="$scriptUrl"></script>
+            @scalajs.html.scripts("client", routes.Assets.versioned(_).toString, name => getClass.getResource(s"/public/$$name") != null)
```
julienrf, line 16: Shouldn’t we do this?

```suggestion
            @scalajs.html.scripts("client", routes.Assets.versioned(_).toString, name => getClass.getResource(s"/public/$name") != null)
```

mn98: That would be a compile error. That code is within an html string, so it appears we need to insert a literal "$" into that line by using $$.

julienrf: I don’t follow. We don’t want the resulting HTML string to contain `name => getClass.getResource(s"/public/$name") != null`, we want to evaluate wether the resource is in our classpath or not.

Am I wrong?

mn98: Hmm, tbh, I'm not sure! :-)
But `$name` is a compile error for sure. In my own very limited experience I've been constructing html in Play using the twirl stuff, like this:

```
@(title: String)(content: Html)(implicit flash: Flash)

<!DOCTYPE html>

<html>
    <head>
        <title>@title</title>
        <link rel="shortcut icon" type="image/png" href="@routes.Assets.versioned("images/favicon.png")">
        <!-- React -->
        <script src="https://unpkg.com/react@@16/umd/react.development.js"></script>
        <script src="https://unpkg.com/react-dom@@16/umd/react-dom.development.js"></script>
    </head>
    <body>
        @flash.get("error")
        @flash.get("success")
        @content
        @scalajs.html.scripts("client", routes.Assets.versioned(_).toString, name => getClass.getResource(s"/public/$name") != null)
    </body>
</html>
```

which would have a single `$` because this is not within a string. I took the passing local test as a green light that I'd done it 'correctly' in that the `$$` resolves to `$` and the script is executed correctly. Am I mistaken? I'm _very_  new to all of this so it's quite possible!

julienrf: I guess the test is insufficient because now I see that we should not use `@` but `$`:

```suggestion
            ${ scalajs.html.scripts("client", routes.Assets.versioned(_).toString, name => getClass.getResource(s"/public/$name") != null) }
```

That one should work.

mn98: That is a compile error too, because the `routes.Assets` is not recognised, presumably because we haven't used twirl's magic `@` anywhere.

julienrf: Ah, it must be a missing import that is usually inserted by default by Twirl. Try with the fully qualified name: `controllers.routes.Assets.versioned`.

mn98: Ok, that compiles and the test appears to pass locally.

julienrf (approved):
Thanks for your help!

Issue #389, opened 2021-01-08
No log for startWebpackDevServer


vhiairrassary:
After `fastOptJS::startWebpackDevServer` succeeded I have no log from webpack-dev-server when recompiling. 

Replacing `logger` usages by `println` fix the problem in:
https://github.com/scalacenter/scalajs-bundler/blob/11cefe060fdcf689796f1e435100ad65525f26fd/sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala#L44-L46

I suspect sbt is "closing" the logger logger when the task is finished, while `fastOptJS::startWebpackDevServer` launches the devServer in a background progress. Any thoughts on the recommended fix? I am more than happy to work on it if needed

My configuration:
- scalajs-bundler: 0.20.0
- scalajs: 1.3.1
- sbt: 1.4.6
- webpack: 4.44.2
- webpack-dev-server: 3.11.1

vhiairrassary:
Logs are here using sbt `1.3.13` but not with sbt `1.4.6`.

The root cause is described in https://github.com/sbt/sbt/releases/tag/v1.4.0:
> Replaces Apache Log4j with our own logger by default to avoid Appender leakage. Use ThisBuild / useLog4J := true to use Log4j.

Rational is available here: https://github.com/sbt/sbt/pull/5731. After using `ThisBuild / useLog4J := true` then logs are available.

Would you rather:
- ask users to use `ThisBuild / useLog4J := true` starting sbt `1.4.0` (with the associated memory leaks)?
- use `println` instead of `logger.info` & `logger.error` in `scalajsbundler.util.Commands`?

Also the decision may sit behind a setting but I think it is not a good idea to expose that for end users.

Anyway, I can prepare a PR for any solution (which would be a code change or a documentation change)

sjrd:
Whatever we do, we shouldn't use `println`, since that is not configurable, and does not support `last`.

vhiairrassary:
👍 Here is a [PoC](https://github.com/scalacenter/scalajs-bundler/pull/391) using sbt global logger. I still need to validate the usage as there is not much documentation about the global logger.

PR #391, opened 2021-01-10
Fix webpack-dev-server logs with sbt 1.4.x

vhiairrassary:
Fixes https://github.com/scalacenter/scalajs-bundler/issues/389

Starting with sbt 1.4.x (see https://github.com/sbt/sbt/pull/5731) it is not possible to use the task logger after the task has been finished. But this is currently how we log outputs from `webpack-dev-server`.

This PR changes the logger to use the [global one](https://github.com/sbt/sbt/blob/aa91a1b304454be0d77c0128c4f1e9f0e5ad0ed8/internal/util-logging/src/main/scala/sbt/internal/util/GlobalLogging.scala#L13-L28) for this particular use case. I still be validate this is the recommended approach as there is not much documentation about it.

Manually tested with sbt `1.2.8`, `1.3.13` and `1.4.6`.


dispalt:
I like this solution, it was frustrating when it went away. +1

cquiroz (approved):
Looks good, thanks

PR #392, opened 2021-01-11
Sync NPM lock file

vhiairrassary:
Npm, as yarn, supports lock files for several versions now so we add the same mechanism to save them.
Feedbacks welcome


vhiairrassary:
Rebased. Let me know if you are interested in this change

cquiroz:
I think we can include this on the RC release

PR #408, opened 2021-08-17
Fix WebpackStats json Reads and update some npm packages

giabao:
Pls see the commit messages

### DIFF
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
@@ -2,7 +2,6 @@
 package scalajsbundler
 
 import play.api.libs.json._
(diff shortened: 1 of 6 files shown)


sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
```diff
@@ -50,18 +49,18 @@ object Stats {
 
   }
 
-  final case class WebpackError(moduleName: String, message: String, loc: String)
+  final case class WebpackError(moduleName: Option[String], message: String, loc: String)
```
evbo, line 52: the docs are a little unclear, but besides `message` it sounds like "various" other properties means perhaps `loc` should be optional too?:
https://webpack.js.org/api/stats/#errors-and-warnings

Issue #402, opened 2021-03-30
Cannot download plug in --- instructions out of date / missing a step?


ScalaWilliam:
On this page, https://scalacenter.github.io/scalajs-bundler/ the instruction says to use:

```scala
addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0d8812f9")
```

However upon following it, I get this:
```
[error] (update) sbt.librarymanagement.ResolveException: Error downloading ch.epfl.scala:sbt-scalajs-bundler;sbtVersion=1.0;scalaVersion=2.12:0d8812f9
[error]   Not found
[error]   Not found
[error]   not found: https://repo1.maven.org/maven2/ch/epfl/scala/sbt-scalajs-bundler_2.12_1.0/0d8812f9/sbt-scalajs-bundler-0d8812f9.pom
[error]   not found: https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/ch.epfl.scala/sbt-scalajs-bundler/scala_2.12/sbt_1.0/0d8812f9/ivys/ivy.xml
[error]   not found: https://repo.typesafe.com/typesafe/ivy-releases/ch.epfl.scala/sbt-scalajs-bundler/scala_2.12/sbt_1.0/0d8812f9/ivys/ivy.xml
[warn] Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? (default: r)
```

sjrd:
Sorry. The correct version is 0.20.0. I need to fix the website.

ansvonwa:
Similar issue: The website now recommends `"0.0.0+1-d083e6df-SNAPSHOT"`

Feel-ix-343:
*hours later

thanks

alekmeckaroski:
This is still an issue, the webpage is not updated with the proper version

PR #413, opened 2022-02-20
update version on install instructions to 0.20.0

fiatjaf:
fixes https://github.com/scalacenter/scalajs-bundler/issues/402

### DIFF
--- a/getting-started.html
+++ b/getting-started.html
@@ -66,9 +66,9 @@
 <p>Add the <code class="hljs">sbt-scalajs-bundler</code> plugin to your Scala.js project, in your <code class="hljs">project/plugins.sbt</code> file:</p>
 <div class="row"><div class="a_linked small-expand columns a_xscroll a_codeblock">
 <pre class="hljs"><code class="language-scala"><span class="hljs-comment">// For Scala.js 1.x</span>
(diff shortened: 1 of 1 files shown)


sjrd:
I could merge this, but it would be overwritten by a bot on the next merge on `main`. The actual source for this is at
https://github.com/scalacenter/scalajs-bundler/blob/b37024143b5a194de11a4395b367711eec2a3248/manual/src/ornate/index.md?plain=1#L11-L14
We should probably replace the dynamic `{{version}}`s by hard-corded ones.

fiatjaf:
Why is the `{{version}}` printing the wrong value though? It would work if `{{version}}` corresponded to the latest release.

sjrd:
I don't know. It's possible that the value of `{{version}}` comes from
https://github.com/scalacenter/scalajs-bundler/blob/d083e6df3fe2ac795ae16ba69696039bf85f85a3/manual/src/ornate.conf#L17
However, I don't know where the `${version}` comes from. :man_shrugging:

fiatjaf:
ahahah, well, apparently it is a built-in thing from ornate, and there seems to be also a [`${tag}`](https://github.com/szeiger/ornate/blob/bfe49970a78bcc834e6555f0e5680aa70086c7ae/doc/ornate.conf#L35) value.

I should try to rebuild the pages using that and send a proper PR, but I'm just trying to learn the basic of basics of Scala here and all this sbt thing is imploding my head so I'll refrain from doing it now :P

Issue #417, opened 2022-05-01
Support `packageManager` in package.json


ptrdom:
https://nodejs.org/api/packages.html#packagemanager

I would see it as an upgrade to `useYarn` setting. Instead, we could could have an `packageManager` setting, which defaults to `npm`, but can be set to `yarn` or `pnpm` with a specified version. Very useful part about this is that npm will automatically download supported package managers.

PR #420, opened 2022-05-05
Package manager abstraction

ptrdom:
References https://github.com/scalacenter/scalajs-bundler/issues/417.

Main idea is to make a package manager abstraction that allows a flexible interaction with plugin's internals so that user's can extend the behavior however they wish.

TODO:
- [x] Implement `pnpm`.
- [x] Fix scripted tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,12 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 16.14.2
+      - name: Enable Corepack
+        run: corepack enable
       - name: Setup yarn
-        run: npm install -g yarn@1.22.15
+        run: corepack prepare yarn@1.22.15 --activate
+      - name: Setup pnpm
+        run: corepack prepare pnpm@7.0.1 --activate
       - name: Unit tests
         run: sbt test
       - name: Scripted tests
--- a/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageManager.scala
+++ b/sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageManager.scala
@@ -1,0 +1,252 @@
+package scalajsbundler
+
+import java.io.File
+
+import sbt._
+import scalajsbundler.util.Commands
+import scalajsbundler.util.JSON
+
+trait PackageManager {
+
+  val name: String
+
+  /**
+    * Runs the command `cmd`
+    * @param args Command arguments
+    * @param workingDir Working directory of the process
+    * @param logger Logger
+    */
+  def run(args: String*)(workingDir: File, logger: Logger): Unit
+
+  def install(baseDir: File, installDir: File, logger: Logger): Unit
+
+  val packageJsonContents: Map[String, JSON]
+}
+
+object PackageManager {
+
+  abstract class ExternalProcess(
+    val name: String,
+    val installCommand: String,
+    val installArgs: Seq[String]
+  ) extends PackageManager {
+
+    def run(args: String*)(workingDir: File, logger: Logger): Unit =
+      Commands.run(cmd ++: args, workingDir, logger)
+
+    private val cmd = sys.props("os.name").toLowerCase match {
+      case os if os.contains("win") => Seq("cmd", "/c", name)
+      case _                        => Seq(name)
+    }
+
+    def install(baseDir: File, installDir: File, logger: Logger): Unit = {
+      this match {
+        case lfs: LockFileSupport =>
+          lfs.lockFileRead(baseDir, installDir, logger)
+        case _ =>
+          ()
+      }
+
+      run(installCommand +: installArgs: _*)(installDir, logger)
+
+      this match {
+        case lfs: LockFileSupport =>
+          lfs.lockFileWrite(baseDir, installDir, logger)
+        case _ =>
+          ()
+      }
+    }
+  }
+
+  trait AddPackagesSupport { this: PackageManager =>
+
+    val addPackagesCommand: String
+    val addPackagesArgs: Seq[String]
+
+    /**
+      * Locally install NPM packages
+      *
+      * @param baseDir The (sub-)project directory which contains yarn.lock
+      * @param installDir The directory in which to install the packages
+      * @param logger sbt logger
+      * @param npmPackages Packages to install (e.g. "webpack", "webpack@2.2.1")
+      */
+    def addPackages(baseDir: File,
+                    installDir: File,
+                    logger: Logger,
+                   )(npmPackages: String*): Unit = {
+      this match {
+        case lfs: LockFileSupport  =>
+          lfs.lockFileRead(baseDir, installDir, logger)
+        case _ =>
+          ()
+      }
+
+      run(addPackagesCommand +: (addPackagesArgs ++ npmPackages): _*)(installDir, logger)
+
+      this match {
+        case lfs: LockFileSupport  =>
+          lfs.lockFileWrite(baseDir, installDir, logger)
+        case _ =>
+          ()
+      }
+    }
+  }
+
+  trait LockFileSupport {
+    val lockFileName: String
+
+    def lockFileRead(
+                      baseDir: File,
+                      installDir: File,
+                      logger: Logger
+                    ): Unit = {
+      val sourceLockFile = baseDir / lockFileName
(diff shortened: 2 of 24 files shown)


ptrdom:
@julienrf @sjrd comments would be appreciated. Could I get some permissions to run CI at will, scripted tests are kind of clunky on my local machine. I could fix up tests that were broken prior to this PR too, but I will need to edit actual CI scripts because I have a feeling it needs some tinkering with Node/npm versions. I do not see where those `AppVeyor` scripts are located, but do we even need them with GitHub Actions being already in place?

ptrdom:
Also, this very much breaks the backwards compatibility, but I assume that it is fine since the plugin is not v1 yet 😅 Let me know if that is not alright, there is definitely a way to stay backwards compatible, although the old design should be deprecated eventually - things like having both npm and yarn settings available simultaneously is obvious clunky.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageManager.scala
```diff
@@ -0,0 +1,163 @@
...
+    val targetLockFile = installDir / lockFileName
+
+    if (targetLockFile.exists()) {
+      logger.debug("Wrote lockfile to " + sourceLockFile)
+      IO.copyFile(targetLockFile, sourceLockFile)
+    }
+  }
+}
+
+case class Npm(
```
sjrd, line 119: Never use `case class`es for things that you don't intend to `match/case` on. These are impossible to evolve in binary-compatible ways.

Use a `final class` with a `Config` parameter instead, where `Config` follows the pattern that you can see at
https://github.com/scala-js/scala-js-js-envs/blob/9d46d49310619de3d722b5d3867fa7599b1a5831/nodejs-env/src/main/scala/org/scalajs/jsenv/nodejs/NodeJSEnv.scala#L213-L270
Yes, it is a pain to write, but it is the only way not to completely lock this down for the future.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageManager.scala
```diff
@@ -0,0 +1,163 @@
...
+    with LockFileSupport
+    with AddPackagesSupport {
+  override val packageJsonContents: Map[String, JSON] =
+    version.map(v => Map("packageManager" -> JSON.str(s"$name@$v"))).getOrElse(Map.empty)
+}
+object Yarn {
+  val DefaultArgs: Seq[String] = Seq("--non-interactive", "--mutex", "network")
+}
+
+case class Pnpm(
```
sjrd, line 150: Since we are using an OOP approach, this is actually something we don't have to back inside scalajs-bundler! Instead, users who want to use `pnpm` can define something like this in their build. We can document this, and then users won't have to request support for other managers in the future, as they'll be able to define whichever they need themselves. This is similar to the design of `JSEnv`s in Scala.js core.

We'd have to keep `Npm` and `Yarn` in scalajs-bundler for compatibility reasons, but there is no reason to *add* more.

It doesn't mean we can't have a *test* with pnpm. On the contrary, such a test ensures that it is indeed possible to do this in user space.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
```diff
@@ -21,12 +22,10 @@ object NpmUpdateTasks {
   def npmUpdate(baseDir: File,
```
sjrd, line 22: We should keep a deprecated overload of this method for compatibility reasons, which will delegate to this one.

Also the Scaladoc needs updating.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/NpmUpdateTasks.scala
```diff
@@ -45,18 +44,16 @@ object NpmUpdateTasks {
   def npmInstallDependencies(baseDir: File,
```
sjrd, line 44: Same here.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/sbtplugin/PackageJsonTasks.scala
```diff
@@ -31,7 +31,8 @@ object PackageJsonTasks {
     webpackVersion: String,
     webpackDevServerVersion: String,
     webpackCliVersion: String,
-    streams: Keys.TaskStreams
+    streams: Keys.TaskStreams,
+    packageManager: PackageManager
   ): BundlerFile.PackageJson = {
```
sjrd, line 36: Same here, and for the other methods below. We should keep deprecated aliases.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/npm/test
```diff
@@ -0,0 +1,2 @@
+> fastOptJS::webpack
+$ exists package-lock.json
```
sjrd, line 2: Missing EOL.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/yarn-interactive/test
```diff
@@ -1 +1,2 @@
 > fastOptJS::webpack
+$ exists yarn.lock
```
sjrd, line 2: Missing EOL.

sjrd (changes requested):

ptrdom:
@sjrd Thank you for a very thorough review! I will make the changes as requested.

.github/workflows/ci.yml
```diff
@@ -21,8 +21,12 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 16.14.2
+      - name: Enable Corepack
+        run: corepack enable
       - name: Setup yarn
-        run: npm install -g yarn@1.22.15
+        run: corepack prepare yarn@1.22.15 --activate
+      - name: Setup pnpm
+        run: corepack prepare pnpm@7.0.1 --activate
```
ptrdom, line 29: Managing package manager installations inside and outside Corepack do not mix well, so we got to do it with Corepack only.

ptrdom:
@sjrd I think I made all of the modifications you requested, PR is ready for a review. I also updated docs.

sbt-scalajs-bundler/src/main/scala/scalajsbundler/PackageManager.scala
```diff
@@ -0,0 +1,252 @@
...
+        case lfs: LockFileSupport  =>
+          lfs.lockFileWrite(baseDir, installDir, logger)
+        case _ =>
+          ()
+      }
+    }
+  }
+
+  trait LockFileSupport {
+    val lockFileName: String
```
ptrdom, line 97: It would be better to convert this to a `File` to allow for better lockfile location customization.

ptrdom: Though I do not have the best idea on where to put `PackageManager.Npm` and `PackageManager.Yarn` so that `baseDirectory` is available during construction.

mims-github:
@ptrdom Thanks for this PR, would love to use pnpm with Scala.js. :)

Anything missing for this PR to be merged?

nadenf:
@sjrd  .. Can we request a review of this. I would be happy to add bun support as well on top of this.

Either way the JS community has much faster options than NPM right now which we should take advantage of.

ptrdom:
I think this PR was ignored for long enough to assume that maintainers are not accepting any PRs to actually develop features, which is fine, but it would have been considerate towards community to explicitly state that.

My suggestions for plugin users who want a package manager abstraction:
- Fork `scalajs-bundler` and publish a community fork.
- Move on to Vite in officially promoted way:
  - https://www.scala-js.org/doc/tutorial/scalajs-vite.html
- Move on to one of the plugins I worked on:
  - https://github.com/ptrdom/scalajs-vite for Vite, feature complete, but I am not working on it any further since I personally think that esbuild is a better bundler for Scala.js.
  - https://github.com/ptrdom/scalajs-esbuild for esbuild, feature complete for web, still working on Electron plugin, and lacks docs, so I would still call it WIP.

Issue #423, opened 2022-05-11
Error parsing webpack stats output


ptrdom:
`fullOptJS::webpack` threw the following error for https://github.com/ptrdom/akka-http-slinky-endpoints4s: 
```
[error] Error parsing webpack stats output
[error] /warnings(0)/moduleName: JsonValidationError(List(error.path.missing),WrappedArray())
[error] /warnings(1)/moduleName: JsonValidationError(List(error.path.missing),WrappedArray())
[error] /warnings(2)/moduleName: JsonValidationError(List(error.path.missing),WrappedArray())
```
JSON output was:
```
"warnings":[{"message":"asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).\nThis can impact web perform
ance.\nAssets: \n  clientprod-opt.109ea631b1f71d35c84e.bundle.js (622 KiB)","stack":"AssetsOverSizeLimitWarning: asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).\nThis can impact web performance.\nAssets: \n  clientprod-opt.109ea
631b1f71d35c84e.bundle.js (622 KiB)\n    at C:\\Users\\Domantas\\IdeaProjects\\open-source\\akka-http-slinky-endpoints4s\\client\\prod\\target\\scala-2.13\\scalajs-bundler\\main\\node_modules\\webpack\\lib\\performance\\SizeLimitsPlugin.js:134:7\n    at Hook.eval [a
s callAsync] (eval at create (C:\\Users\\Domantas\\IdeaProjects\\open-source\\akka-http-slinky-endpoints4s\\client\\prod\\target\\scala-2.13\\scalajs-bundler\\main\\node_modules\\tapable\\lib\\HookCodeFactory.js:33:10), <anonymous>:9:1)\n    at Hook.CALL_ASYNC_DELEG
ATE [as _callAsync] (C:\\Users\\Domantas\\IdeaProjects\\open-source\\akka-http-slinky-endpoints4s\\client\\prod\\target\\scala-2.13\\scalajs-bundler\\main\\node_modules\\tapable\\lib\\Hook.js:18:14)\n    at C:\\Users\\Domantas\\IdeaProjects\\open-source\\akka-http-s
[...]

Issue #338, opened 2020-05-15
What is the significance of stats failing and how to prevent?


evbo:
Webpack appears to correctly build an electron app that I can run with `electron`, but at the very end I see the message: 
```
[info] Writing scalajs.webpack.config.js
[error] Error parsing webpack stats output
[error] /assets: JsonValidationError(List(error.path.missing),WrappedArray())
[error] /time: JsonValidationError(List(error.path.missing),WrappedArray())
```

I see in the change log this is expected if any std output gets generated:
https://github.com/scalacenter/scalajs-bundler/blob/08cd66ba26716301c09cae5853b42a7cc31e1c05/manual/src/ornate/changelog.md

So is it a concern? Any thoughts on how I can silence std output to avoid this error?

cquiroz:
the plugin will read the stats produced by webpack to identify the assets it built. The error maybe an indication that the format changed.
What version of webpack are you using?

evbo:
Sorry should have included that:

> Plugins:
addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.0.0")
addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0.17.0")

> Settings:
version in webpack := "4.39.1",
version in startWebpackDevServer := "3.7.2",

I also just tried these versions and I get the same error:
> version in webpack := "4.43.0",
    version in startWebpackDevServer := "3.11.0",


thanks for the quick response. Just to clearly emphasize, this is only an issue for electron, whereby the best practice (no official documentation on this that I could find) seems to be configuring a _multi-compiler_ like so:
https://github.com/tarunlalwani/electron-webpack-typescript-boilerplate/blob/master/webpack.config.js
https://www.sitepen.com/blog/getting-started-with-electron-typescript-react-and-webpack/

If I just make a conventional config, an object not an array of objects, I do not see any errors.

evbo:
Any ideas how to debug this issue? One unfortunate side effect is that the exit code causes jenkins builds to fail even though the build is actually successful.

Thanks

evbo:
Here's an example of it failing (now with bundler: `0.20.0` and same webpack as before). Even though I had a legitimate error with a JSImport referencing a ".conf" file that should have been a ".json" file, the error seems to be getting swallowed by this instead (no errors were shown prior in the build):  

[...]

PR #424, opened 2022-05-13
Failure to parse Webpack stats fails the build

ptrdom:
Failure to parse Webpack stats can fail builds with certain Webpack configs. It should probably be a critical failure, rather than logging with return of an `Option`.


ptrdom:
Related to https://github.com/scalacenter/scalajs-bundler/issues/423.

evbo:
and related to https://github.com/scalacenter/scalajs-bundler/issues/338 which was closed because I presumed the PR would get merged, but it never did - shame on me :)

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
```diff
@@ -115,13 +115,13 @@ object Stats {
   )(Asset.apply _)
 
   implicit val errorReads: Reads[WebpackError] = (
```
evbo, line 117: @ptrdom what do you think of just using semi-auto derivation in `Stats.scala`?:

```scala
  implicit val assetsReads: Reads[Asset] = Json.reads[Asset]
  implicit val errorReads: Reads[WebpackError] = Json.reads[WebpackError]
  implicit val warningReads: Reads[WebpackWarning] = Json.reads[WebpackWarning]
  implicit val pathReads: Reads[Path] = Reads.StringReads.map(new File(_).toPath)
  implicit val statsReads: Reads[WebpackStats] = Json.using[Json.WithDefaultValues].reads[WebpackStats]
```

I saw this used in https://github.com/scalacenter/scalajs-bundler/pull/408 and tested this works and saves a few lines of code ;)

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
```diff
@@ -50,9 +50,9 @@ object Stats {
 
   }
 
-  final case class WebpackError(moduleName: String, message: String, loc: String)
+  final case class WebpackError(moduleName: Option[String], message: String, loc: String)
```
evbo, line 53: the docs are a little unclear, but besides `message` it sounds like "various" other properties means perhaps `loc` should be optional too?:
https://webpack.js.org/api/stats/#errors-and-warnings

evbo:
Also, @ptrdom did you try `sbt publishLocal` with your fork? I tried it and my project failed to resolve it as a dependency due to `scalajs-bundler-linker` not getting published locally. I had to add it to the build.sbt aggregate statement here:
https://github.com/scalacenter/scalajs-bundler/blob/main/build.sbt#L90

Curious to know how you managed to publish locally without that fix...

ptrdom:
@evbo I ran into the same issue and simply published the module explicitly, same as the CI workflow does it - https://github.com/scalacenter/scalajs-bundler/blob/main/.github/workflows/release.yml#L21-L22. Adding the project to aggregate makes sense to me.

ptrdom:
@sjrd Can we merge this?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
```diff
@@ -107,32 +141,10 @@ object Stats {
...
-    (JsPath \ "outputPath").readNullable[String].map(x => x.map(new File(_).toPath)) and // It seems webpack 2 doesn't produce outputPath
-    (JsPath \ "errors").read[List[WebpackError]] and
-    (JsPath \ "warnings").read[List[WebpackWarning]] and
-    (JsPath \ "assets").read[List[Asset]]
-  )(WebpackStats.apply _)
+  implicit val assetsReads: Reads[Asset] = Json.reads[Asset]
+  implicit val errorReads: Reads[WebpackError] = Json.reads[WebpackError]
+  implicit val warningReads: Reads[WebpackWarning] = Json.reads[WebpackWarning]
+  implicit val pathReads: Reads[Path] = Reads.StringReads.map(new File(_).toPath)
+  implicit val statsReads: Reads[WebpackStats] = Json.using[Json.WithDefaultValues].reads[WebpackStats]
```
sjrd, line 148: These changes don't seem to have anything to do with the purpose of this PR, have they?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/BundlerFile.scala
```diff
@@ -88,14 +85,36 @@ object BundlerFile {
...
+            if (asset.exists()) {
+              (notExisting, existing :+ asset)
+            } else {
+              (notExisting :+ asset, existing)
+            }
+        }
+      if (notExisting.nonEmpty) {
+        throw new RuntimeException(
+          s"Webpack failed to create application assets:\n" +
+            s"${notExisting.map(asset => s"${asset.getAbsolutePath}").mkString("\n")}")
```
sjrd, line 113: I you're going to throw anyway, this method can be simplified a lot:
```scala
val assets = stats.resolveAllAssets(targetDir)
val nonExisting = assets.filter(!_.exists())
if (nonExisting.nonEmpty) {
  throw ...
}
assets
```

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Webpack.scala
```diff
@@ -279,11 +273,16 @@ object Webpack {
...
-  def run(nodeArgs: String*)(args: String*)(workingDir: File, log: Logger): Option[WebpackStats] = {
+  def run(nodeArgs: String*)(args: String*)(workingDir: File, log: Logger): WebpackStats = {
     val webpackBin = workingDir / "node_modules" / "webpack" / "bin" / "webpack"
     val params = nodeArgs ++ Seq(webpackBin.absolutePath, "--profile", "--json") ++ args
     val cmd = "node" +: params
-    Commands.run(cmd, workingDir, log, jsonOutput(cmd, log)).fold(sys.error, _.flatten)
+    log.debug(s"Running command [${cmd.mkString(" ")}]")
+    Commands.run(cmd, workingDir, log, jsonOutput(cmd, log))
+      .fold(
+        sys.error,
```
sjrd, line 283: If you're going to throw when the `Try` is a `Failure`, you as well do
```scala
val runResult = Commands.run(...).get
```
But then again, why change `Commands.run` to return a `Try` in that case, instead of directly throwing the exception?

sbt-scalajs-bundler/src/main/scala/scalajsbundler/util/Commands.scala
```diff
@@ -1,21 +1,23 @@
 package scalajsbundler.util
 
 import sbt.Logger
-import java.io.{InputStream, File}
+import java.io.{File, InputStream}
+
 import scala.sys.process.Process
```
sjrd, line 6: Spurious changes.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-stats/test
```diff
@@ -0,0 +1 @@
+> fullOptJS::webpack
```
sjrd, line 1: I don't understand how this test relates to the fix. If this PR is about failing a build, there should a be at least one negative test line (starting with `-`).

sjrd (changes requested):

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
evbo: @sjrd It does because you need to update the JSON decoders to parse the new `Option` types that have been added. So rather than explicitly modify the individual fields it parses as Options, this solves the problem using derived decoders, which is also fewer lines and I don't think there are any downsides since it's compile-time derivation (maybe slightly longer time taken compiling?)

build.sbt
```diff
@@ -87,7 +87,7 @@ val manual =
 val `scalajs-bundler` =
   project.in(file("."))
     .settings(noPublishSettings: _*)
-    .aggregate(`sbt-scalajs-bundler`, `sbt-web-scalajs-bundler`)
+    .aggregate(`sbt-scalajs-bundler`, `sbt-web-scalajs-bundler`, `scalajs-bundler-linker`)
```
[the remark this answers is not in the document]
evbo: that was my bad - whoops. See comment: https://github.com/scalacenter/scalajs-bundler/pull/424#issuecomment-1126867086

Note: I am currently using scalajs-bundler published locally with that aggregate and it has been running fine, though I'm not sure of the implications to a different "cross-version setup".

sbt-scalajs-bundler/src/main/scala/scalajsbundler/Stats.scala
ptrdom: Since the model has changed, `Reads` do have to be modified, and derived decoders are a simplification of those changes.

sbt-scalajs-bundler/src/sbt-test/sbt-scalajs-bundler/webpack-stats/test
ptrdom: This test would fail in current `main` branch, the changes added introduce fixes that actually make it pass. Is that fine?

ptrdom:
@sjrd I made changes for which I did not have further questions, please review the discussions and the code when you can.

annedroiid:
@sjrd Are the latest changes acceptable? Do you need any help with this?

rmmeans:
Wondering if this PR will be up for consideration in the future? I know many are moving to Vite, but our current build process is causing problems with Vite since we're packaging node modules for browser (Autoprefixer and postcss). Didn't have much luck working through (tried following advice https://github.com/postcss/postcss/issues/830, but wasn't able to get it to work). 

Was hoping to upgrade to webpack5 on scalajs-bundler until we can refactor how we do CSS to not need autoprefixer running in the browser and then switch to Vite.

PR #422, opened 2022-05-09
Drop AppVeyor

ptrdom:
With CI being already migrated to GitHub Actions, keeping AppVeyor seems unnecessary. Less CI scripts to maintain. Since GitHub Actions require approval for first time contributors and AppVeyor did not, maybe it would be good to reconsider loosening this rule for GitHub Actions in this repository.

PR #425, opened 2022-05-14
Fix AppVeyor

ptrdom:
Until we figure out what is up with https://github.com/scalacenter/scalajs-bundler/pull/422, it would be great to have AppVeyor CI actually passing.


sjrd (approved):
Thanks!

Issue #415, opened 2022-03-30
Documentation reports incorrect version


Quafadas:
I think that the front of this page should be; 
https://scalacenter.github.io/scalajs-bundler/

addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0.20.0")

rather than 

addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0.0.0+1-b3702414-SNAPSHOT")

ptrdom:
Looking at CI setup, it should fix itself up upon next commit to main, but we should probably only publish to GitHub Pages only on tags, rather than every commit to main - https://github.com/scalacenter/scalajs-bundler/blob/main/.github/workflows/ghpages.yml#L7.

OndrejSpanel:
I agree. The documentation should reference a stable version, not a SNAPSHOT one. At the moment this is perhaps:

```
addSbtPlugin("ch.epfl.scala" % "sbt-scalajs-bundler" % "0.21.0")
```

The #421 claims to fix this, but still there is a snapshot version listed at https://scalacenter.github.io/scalajs-bundler/

PR #426, opened 2022-05-14
Fix release version numbering

ptrdom:
`fetch-depth: 0` is required for `checkout` action to fetch tags, which are needed to correctly set versions. All releases from the workflow have been `0.0.0` so far - https://oss.sonatype.org/content/repositories/snapshots/ch/epfl/scala/scalajs-bundler-linker_2.12/.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          # fetches all tags, required to correctly set the version
+          fetch-depth: 0
       - uses: olafurpg/setup-scala@v12
       - uses: olafurpg/setup-gpg@v3
       - uses: coursier/cache-action@v5


ptrdom:
Resolves https://github.com/scalacenter/scalajs-bundler/issues/415.

ptrdom:
@sjrd Can we merge this?

.github/workflows/release.yml
```diff
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
sjrd, line 12: Could you add a comment that this is required to fetch tags? Because it's completely non-obvious.

sjrd (changes requested):

.github/workflows/release.yml
ptrdom: While it is a documented behavior of the action, I agree that it is quite unexpected 😅 I added a comment as requested.

sjrd (approved):

